Issue #41, opened 2019-01-28
i2c option to turn on hardware pullups


trentgill:
if crow is the only master on the bus, and no powered busboard provided, we need i2c pullup in order for things to work.

`ii.pullup( enable )`

Issue #29, opened 2018-12-24
ii autoloads module's get/set functions as needed


trentgill:
the initial version requires the user to 'require' the table for their module to be remote-controlled:
`require 'JF'`

the design would mean a call to eg: `ii.jf.transpose()` would fail if the above hadn't been required.

we fundamentally require this now because their is a hard RAM limit and loading all of the helper for all of the potential i2c clients would use far too much.

instead, we can hide this limitation from the user, making it appear as though all of the tables are always loaded. whenever there is a call to ii.jf.___ and the function is not found, we lookup if the jf table is nil (if it's not, then it suggests a typo), but if it is, we first `require 'JF'` before executing the code.

this means that the RAM is only used after the user first uses a function from that subset even though it appears as though all the functions are always available.

//

note: this requires ii.<module>.listcommands() to be implemented in C, printing a c-string (from a header file) as the contents. or perhaps this string/array is passed back to lua for printing/matching.

tehn:
i think it's really smart to not preload everything into RAM.

but i'm starting to prefer less hidden-magic syntax fixes in the norns scripting system. feeling it's ok to have scripts do more explicit `requires` and i see runtime ramifications with auto-loading (ie, first ii command gets lagged/delayed waiting for the lib to preload)

trentgill:
my thinking here is that this hidden-magic is not a syntax fix, but rather a workaround for RAM limitations. the goal was to have the II.<module> files work as if they are all loaded already. in general we don't use `require` anywhere except in the standard lib (`crowlib`) itself (well `dofile` actually), so in a way i like the idea that crow scripts never deal with external dependencies - everything is just always available.

i take the point that it is kind of a hack though, and for sure there will be some runtime delay for that first command, though i'd guess it's negligible IRL. it's already working in PR #69 but i can remove if we decide it's better without it.

tehn:
by all means go on ahead--- i was simply trying to set priorities, so if _not_ implementing this actually creates just as much (or more) work, it'd be better to get it in there.

i'm less worried about the delay.

PR #69, opened 2019-02-26
I2C events, pullups & helpers

trentgill:
`II.pullup(state)`
Turns on STM32 i2c pullups. These resistors are 40k each so somewhat out of spec [2k2, 10k], but thankfully on the too-weak side. In practice turning them on made communication with TXi & W/ work though was not stress tested. This capability could be activated on W/ as well.
Solves #41.

`II.<module>.help()`
Prints a list of the devices supported functions (as far as crow is aware). This help list is autogenerated from the i2c descriptor file (in lua/ii/*.lua), and is directly copy&paste-able prototypes.

`II.<module>.event()`
The event handler sends a `^^` message to the remote-host. Thus norns or another host should match on:
`II.<module>(event,value)`
Note the allcaps `II`. It's unclear to me if the dot-notation is the right way to do namespacing in the remote context.

**Autoload II submodules as needed**
When a module is first used, but not currently in RAM, it will `dofile()` the library & save to the `II` table. This means the user can interact as if all module's II tables are loaded, but we only suffer the RAM hit when the module is actually invoked. The modules are stored in Flash as C-strings like everything else.
Solves #29.

**README**
Updated to reflect that all calls to `II` must be capitalized. This could be changed in future, but for now it works so I'm just making it all consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -553,6 +553,11 @@ end
 
 ```
 
+Note that by default the `event()` defaults to a satellite action which is
+`^^II.<module>(<event>,<value>)`
+for eg:
+`^^II.txi('value', 0.1)`
+
 ## Calibration
 
 crow has an in-built calibration mechansim to allow the inputs and outputs to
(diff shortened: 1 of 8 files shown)


trentgill:
Retracting. will resubmit with i2c follower implementation added momentarily.

PR #70, opened 2019-02-26
print now automatically attempt to stringify non-string types

trentgill:
like normal lua `print`.
allows tables to be printed which will just show their memory address.


tehn:
excellent!

PR #85, opened 2019-03-03
I2C as leader, pullups, helpers, autoloading, documentation

### LINKED ISSUE #29: ii autoloads module's get/set functions as needed

### LINKED ISSUE #41: i2c option to turn on hardware pullups

trentgill:
I2C implementation for crow-as-leader. Usage is as described (and updated) in the README.md

crow-as-follower is still unreliable though it is unclear whether the problem is isolated in the crow-to-crow case, or a more general problem with the HAL library usage. This will be able to be tested after TT support is added.

`II.pullup(state)`
Turns on STM32 i2c pullups. These resistors are 40k each so somewhat out of spec [2k2, 10k], but thankfully on the too-weak side. In practice turning them on made communication with TXi & W/ work though was not stress tested. This capability could be activated on W/ as well.
fixes #41.

`II.<module>.help()`
Prints a list of the devices supported functions (as far as crow is aware). This help list is autogenerated from the i2c descriptor file (in lua/ii/*.lua), and is directly copy&paste-able prototypes.

`II.<module>.event()`
The event handler sends a `^^` message to the remote-host. Thus norns or another host should match on:
`II.<module>(event,value)`
Note the allcaps `II`. It's unclear to me if the dot-notation is the right way to do namespacing in the remote context.

**Autoload II submodules as needed**
When a module is first used, but not currently in RAM, it will `dofile()` the library & save to the `II` table. This means the user can interact as if all module's II tables are loaded, but we only suffer the RAM hit when the module is actually invoked. The modules are stored in Flash as C-strings like everything else.
fixes #29.

**README**
Updated to reflect that all calls to `II` must be capitalized. This could be changed in future, but for now it works so I'm just making it all consistent.


trentgill:
i've tested with TXi and JF and they are communicating reliably. there is an issue with TXi behaviour where it only returns the first channel, but I think that might be because I have misunderstood it's communication protocol (it seems to have some weird bitmasking happening). But that seems outside the scope of this PR. importantly the communication itself is reliable & correctly framed.

uppercase is because the calls are dealing with the module itself, not an object created by the module. all the other modules (Input, Output, Asl) start with uppercase,  it's only the objects created by those modules that are lowercase. it's weird i know. basically it's going to be inconsistent somewhere, and for now i chose to keep the module style consistent (using capitals inside the modules). we could simply add an alias in crowlib.lua `ii = II` but again i'll leave that for another PR.

Issue #77, opened 2019-02-28
Crashes with high throughput over TTY


trentgill:
A <50ms input 'stream' will cause either the host program (eg Max) or crow itself to crash.

Potential causes:
- TTY input buffer overflowing (should discard later data?)
- TTY output buffer overflowing (should reject message?)
- single message chunks being too large
- Max [serial] object being polled too slowly (or quickly?)

ngwese:
I’ve experienced this. I am wondering if the Max crash is simple a casualty of the crow crash and the associated serial device disappearing (and Max’s serial object not handling that gracefully)

Issue #20, opened 2018-11-27
Max4Live ->crow object


trentgill:
jason needs it.

trentgill:
Should be a MIDI Effect so the MIDI TRS input can be used to play an Ableton instrument.

whimsical-sam:
Saying hi for trent!

trentgill:
- [ ] Generic usb-port autoselection
OSX ports are of the form: `usbmodem14*1` and windows `COM*`

PR #92, opened 2019-03-07
TTY reliability update & Max object fixes

trentgill:
**warning** serial port baudrate is changed to 460800 (115200 *4) for higher throughput. needs to be updated in `druid` etc.

LMK if changing the baudrate isn't possible in `druid` and I can roll back.

A bunch of changes / refinements to the TTY driver to help improve reliability. I'm now able to stream 2 channels of Input data every 5ms with no crashes to Max. Still seeing reliability issues on Windows though.
Improves #77 but leaving open as it doesn't seem finished.

- Increase USB baudrate for higher-throughput. Reliability seems stable.
- USB Transfer (crow->host) is now on 5ms period. Could likely still go faster
- Max [serial] object matches crow baudrate & faster polling (tried down to 2ms with no problem on OSX).
- Case where a `\r` character is lost no longer causes Lua to fail in Max
- USB driver no longer abandons data if the driver isn't yet ready when the interrupt occurs (waits til next period).
- Max object autoconnects to 3 different ports. Improvements requested in #20 
- Avoid automatic line-endings in caw.c being postponed to next USB interrupt

### DIFF
--- a/lib/caw.c
+++ b/lib/caw.c
@@ -25,18 +25,24 @@ void Caw_send_raw( uint8_t* buf, uint32_t len )
 // luachunk expects a \0 terminated string
 void Caw_send_luachunk( char* text )
 {
+    const uint8_t newline[] = "\n\r";
+uint32_t old_primask = __get_PRIMASK();
+__disable_irq();
     USB_tx_enqueue( (uint8_t*)text, strlen(text) );
-    uint8_t newline[] = "\n\r";
-    USB_tx_enqueue( newline, 2 );
+    USB_tx_enqueue( (uint8_t*)newline, 2 );
+__set_PRIMASK( old_primask );
 }
 
 void Caw_send_luaerror( char* error_msg )
 {
-    uint8_t leader[] = "\\";
-    USB_tx_enqueue( leader, 1 );
+    const uint8_t leader[] = "\\";
+    const uint8_t newline[] = "\n\r";
+uint32_t old_primask = __get_PRIMASK();
+__disable_irq();
+    USB_tx_enqueue( (uint8_t*)leader, 1 );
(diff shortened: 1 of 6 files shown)


trentgill:
@tehn yes the baudrate was utterly arbitrary. i'll try 1M now and if successful will open a new PR with update & squash+merge directly.

trentgill:
All looks good here. Connection is solid at 1MHz and from my testing (streaming both channels of inputs, while sending print() messages to crow) seems reliable.

Merging now. At this point things look quite stable on my end, so will figure out how to get everyone updated.

tehn:
good to hear!

ngwese:
heads up that the new fast baud rate presents a challenge for the `druid` type apps which use POSIX `termios.h` and `tcsetattr(...)` calls to set the speed of the serial line... it looks like the speeds defined above `B115200` are somewhat arbitrary and differ between linux and macOS.

Linux (ubuntu 18.04):
```
...
#define  B1152000 0010011
#define  B1500000 0010012
#define  B2000000 0010013
#define  B2500000 0010014
#define  B3000000 0010015
#define  B3500000 0010016
#define  B4000000 0010017
```

macOS (10.4)
```
...
#define B115200 115200
#define B230400 230400
```

as it stands it isn't clear if it is possible to support rates above 115200 in a portable fashion.

trentgill:
Happy to revert to 115200. I don't think the baudrate is actually the bottleneck anyway, at least with the scripts I've been writing so far.

Will PR this change back asap if @tehn agrees.

tehn:
reversion is fine.

i'm not convinced the speed is actually limited in hardware, that seems fundamentally arbitrary. for example, i know the FT245 can be set to _any_ baudrate and still be full throughput. but also i may be totally off on this.

Issue #39, opened 2019-01-24
refine calibration engine


trentgill:
- [x] add lua hook to print the current calibration values
- [x] add lua hook to recalibrate the io
- [x] print a warning over TTY saying to make sure no cables attached. also upon success
- [x] add zero-check for second input channel (currently just copied from adc0 )
- [x] test accuracy of v8 once we have real CV examples running
- [x] add bounds checks to confirm the data is correct. return error to lua if not (could be a cable attached)
- [x] as it only happens in factory, increase timing, add delays for settling time, make it more robust

trentgill:
Still work to be done!
As of #156 the measurements are:

Listed are mV delta from expected value at (-3, 0, 3, 7) volts
Input 1:
- Output[1] = (-19,-5,4,14)
- Output[2] = (-20,-5,3,14)
- Output[3] = (-19,-5,4,15)
- Output[4] = (-20,-5,4,15)
Input 2:
- Output[1] = (-7,0,8,17)
- Output[2] = (-8,0,7,17)
- Output[3] = (-7,0,7,18)
- Output[4] = (-2,1,3,7)

As can be seen the 'zeroing' of Input 2 is more accurate than Input 1. Input 2 is near perfect & it's ground reference for offset testing is a direct connection to ground through switched jack. Input 1 on the other hand measures it's ground reference via the MUX508 & a 330r resistor. We can assume the ON resistance of the MUX and 330r are causing the discrepancy and need to be factored out somehow.

The discrepancy in Output[4] across the 2 input channel measurements is confounding. Perhaps due to a measurement error? Perhaps something weird with the routing on the pcb? Seems strange that for input[1] the output[4] measurement is near identical to the rest, but output[4] seems off only on input[2]. These measurements were doubly-confirmed though.

The scaling is clearly not being calculated correctly. For input[1] all outputs have between 33 & 35mV differential between -3 & 7V. This differential is evenly distributed across the 4 measurements, so doesn't seem to be caused by non-linearity, but is just an error in the calculation of the scaling factor. Similarly the input[2] measurements show a similar differential (though only 24-25mV excluding the output[4] outlier). Currently input[2] just copies the scaling factor from input[1] as there is no in-built mechanism to test for this without manually patching. Even though the scaling is incorrect, in practice it still seems far better to copy the inaccurate scaler from input[1] than to use the calculated values (which give >50mV differential).

[...]

PR #96, opened 2019-03-15
Calibrate lib cleanup

trentgill:
`Cal` library is now included by `crowlib.lua` by default. I think this might have got lost in a revert ages ago. Users can call functions as outlined in the README, rather than calling the C-hooks directly.

Added (and fixed incorrect) messages to the user (and debug port) to make it clear when calibration is underway. Prettier-print for `Cal.print()` now, and if default calibration is active it doesn't print the (incorrect) data in flash, but just a 'using defaults' message.

Improves but doesn't complete #39

### DIFF
--- a/lib/flash.c
+++ b/lib/flash.c
@@ -90,7 +90,6 @@ uint8_t Flash_is_calibrated( void )
 
 void Flash_clear_calibration( void )
 {
-    printf("clear user script\n");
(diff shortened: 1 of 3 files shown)


tehn:
this looks incredibly helpful, the calibration process is starting to make sense to me now.

PR #101, opened 2019-04-14
update submodules

tehn:
i saw the submodules were updated, figured we should get in sync? let me know if this is an incorrect assumption.

tested on hardware.


trentgill:
I'm a little confused about how this works as i still really *get* submodules.

Regarding those changes, the wrDsp changes don't relate to crow and they aren't in my local crow repo. That said it won't hurt to update them.

For the DFU, indeed you *do* want those changes. The strange thing to me though is that my main crow repo already thinks it's up to date. Usually if I make a change to a submodule the master repo says changes have been made and asks you to commit (right?). Is there some command other than `git pull` that also updates any submodules that have been updated?

Happy for you to merge this, I just don't want it to cause any conflicts and I don't understand submodules well enough to know if conflicts will arise.

tehn:
i think the trick to submodules is just maintaining sync when desired. it seems that right now the repo is pegged to an old version of the submodules.

i don't think this merge causes any trouble. if it does i'm happy to rebuild the structure.

i can do a clean clone and see if things are ok first.

tehn:
just checked (luckily we now have fiber internet!!)

compiles fine. but indeed the DFU submodule isn't on a branch. when i check out master of DFU it sync's it up. i'd suggest merging this PR.

PR #102, opened 2019-04-14
event system

tehn:
event system framework added (based on libavr32)

currently not hooked up to anything--- events are not generated.

here's what i think we need:

- metro ticks https://github.com/monome/crow/blob/master/lib/metro.c#L89
- i2c callbacks https://github.com/monome/crow/blob/master/lib/ii.c
- adc timer https://github.com/monome/crow/blob/master/lib/io.c#L93
- ASL ?? (i thought there was a callback for end-of-slope but i can't find it?)

anything else?

the USB input processing is already effectively an "event system" itself so there's no need to integrate it.

QUESTION: should all events get processed on each loop? or just one and then jump back to USB input query (caw)? (we can just test this later.)

(aside, what's the tab/space preference?)

### DIFF
--- a/lib/detect.c
+++ b/lib/detect.c
@@ -65,12 +65,11 @@ void Detect( Detect_t* self, float level )
                     self->state = 0;
                     if( self->change.direction != 1 ){ // not 'rising' only
                         //(*self->action)( self->channel, (float)self->state );
-            event_t e;
-            e.type = E_change;
-            e.index = self->channel;
-            e.data = (float)self->state;
-            event_post(&e);
-
+                        event_t e;
+                        e.type = E_change;
+                        e.index = self->channel;
+                        e.data = (float)self->state;
+                        event_post(&e);
                     }
                 }
             } else { // low to high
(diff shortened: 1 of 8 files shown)


trentgill:
Ahh, sorry emailed before seeing this. Some specific responses:

ASL callback is called `toward`.
In general, all Lua callbacks are defined at the end of `lib/lualink.c` and in a nice short list at end of `lib/lualink.h`.

ADC also needs an event for a 'change' detection, rather than a timer-based 'stream'.

As opposed to my email, I agree that the USB handling is already effectively an event system, and there's no possibility for conflicts if the event queue is handled sequentially in the main loop. Big relief!

Again, the i2c callback can be restricted to 'response from a follower', and the crow-as-leader should come later.

trentgill:
Regarding the 'process all' or 'just one' approach, I think we should start with 'just one'. The usb_rx won't happen more often than every 1ms, and the 'is the buffer non-empty?' check is super quick so doesn't compromise much to run the loop skinnier but more often.

Reason for this is we want to make sure the usb_rx is able to override in a context where the lua environment has become unresponsive, so the user can type `^^reset` or similar and have that handled reliably.

tehn:
for `toward` do you mean the event should happen here:

https://github.com/monome/crow/blob/master/lib/lualink.c#L180

ie, is `_go_toward` being called by an interrupt?

tehn:
change added. this is ready to go. stream, change, metro, all tested working on hardware (with `druid`)

sorry about the tab/space. let me know what settings you want and i can scrub it

trentgill:
4-space indentation.
this implementation seems to be working well.

i need to change the slope.c implementation to get `toward()` integrated in the event system. should we merge this now, and then i'll open a separate PR for that integration?

tehn:
i'll reformat then merge.

PR #104, opened 2019-04-17
USB stability improvements

### LINKED ISSUE #77: Crashes with high throughput over TTY

trentgill:
Just 2 changes that make a massive difference to usb stability.

(Almost) Stable test case -- this finally crashed after ~10minutes of abuse:
- Running all 4 channels of ASL as LFOs with 33-133hz cycles
- Streaming both ADC channels at 200Hz (5ms intervals)
nb: There are some inaccuracies of data transmission. Seems caused by crow's tx buffer filling up, and the last message being split across neighbouring transmission packets. This could likely be fixed on the host implementation side.

Changes:
1. The 'pointer' variables into usbrx&tx buffers are clarified to reflect the way the queue is simply resetting. Previously this was a circular buffer, but was stripped back for simplicity, now the code reflects the simpler approach.
2. When pushing data to the usb_tx buffer it will now correctly limit the length of an overflowing message to exactly fill the buffer, discarding trailing characters (should this discard the whole packet?)
3. Fixed the edge case of the usb_rx buffer overflow bug that was mostly fixed in my previous PR.
4. The USB allocator has been customized to force all bytes to 0xFF. This solves a problem where usb transmission to the host would fail after every second soft restart, but not after a cold restart (usb_rx would continue to work regardless). Seemed to be something with an ongoing transfer being incomplete when resetting. Zero'ing the buffer caused it to fail *every* time, but `one-ing` the buffer makes it work reliably after both hard & soft reset. This feels like an error in the USB library, but I preferred this solution as it doesn't require changes to the STM library.

I'm going to say fixes #77 .

### DIFF
--- a/usbd/usbd_cdc_interface.c
+++ b/usbd/usbd_cdc_interface.c
@@ -64,7 +64,6 @@ USBD_CDC_LineCodingTypeDef LineCoding = { 115200  // baud rate
 uint8_t UserRxBuffer[APP_RX_DATA_SIZE];
 uint8_t UserTxBuffer[APP_TX_DATA_SIZE];
 uint32_t UserTxBufPtrIn  = 0;
-uint32_t UserTxBufPtrOut = 0;
 uint32_t UserRxBufPtrIn  = 0;
 
 TIM_HandleTypeDef  USBTimHandle;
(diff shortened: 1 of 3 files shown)


trentgill:
Lastly I'd note that I *assume* the 'after 10 mins' crash is still related to calling into Lua while Lua is running another thread. Thus I still believe we need the event-queue to make the system truly reliable.

That said, this PR (and the previous PRs for GC, usb_rx & tell-in-C), make the system orders of magnitude more stable. The test cases are getting more extreme & the time-to-crash are becoming far longer.

Feels good!

PR #110, opened 2019-04-19
lualink wraps the event_post functions, rather than calling via events.h

trentgill:
small change to better isolate the `detect` library with the intention of it being reusable in other systems.

with these changes the detect library is agnostic to what happens when a threshold change is detected, leaving that decision to the user program.

slopes.c will be similar, but obviously there are other changes that need to be made there!

i'll make a new issue about formally moving these libraries out of the crow repo for once they've crystalized into a finished form.

### DIFF
--- a/lib/detect.c
+++ b/lib/detect.c
@@ -1,5 +1,4 @@
 #include "detect.h"
-#include "events.h"
 
 #include <stdlib.h>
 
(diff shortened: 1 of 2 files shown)


trentgill:
jumped the gun on this - didn't realize that all the callbacks were done the same way.
will be a 20min job later, but i'm trying to get the slopes.c updated first.

Issue #112, opened 2019-04-19
slopes.c timing compensation for event queue approach


trentgill:
3 techniques:

1 set a flag & start counting samples when the `(*self->action)()` is called. when the next `toward()` is called, we can compensate for this delay

2 setup a continuation so if the `toward()` is called *before* the samples have been served to the dac, they can be updated with the next slope. this may be more difficult once the shaper engine is implemented?

3 change boundary checks so `self->action` is run one dsp block *before* the breakpoint. the event loop is run *between* dsp blocks, so this hopefully means the `toward()` will have been called by the time the next sample runs. we update the slopes.c handling of toward to store the new values, which will be executed without a callback when the threshold is crossed. this will actually make it *way* easier to handle sub-sample compensation as there's no opaque state-change happening.

trentgill:
Will be unnecessary with ASL2. Closing

PR #113, opened 2019-04-19
Slopes now callback via the event system. isolate event system.

trentgill:
slopes.c now allows the `(*self->action)()` call to happen asynchronously. just waits at the end of the current slope until the event system calls the next slope via the `toward()` function.
- this introduces 1-32 samples (<1 dsp block) of jittery-delay. i've opened #112 to address this.

events.h is now *only* accessed from lualink.c and main.c. the idea is that the events queue is owned by the lua environment. this arrangement adds minimal overhead, but makes maintenance far simpler (i think?!)
- this adds a `L_queue_*` function beside each `L_handle_*` function in lualink  which just wraps the `event_post` call.

side note: no longer clipping the adc range in the callback to lua. this vals are inherently limited in the driver. i actually think the reason this was here is that it stopped the lua env crashing way back before we realized the crashes were caused by the GC settings.

//

i'm running the same stress test as before and after 30mins have yet to see a crash!

### DIFF
--- a/lib/metro.c
+++ b/lib/metro.c
@@ -6,7 +6,6 @@
 #include "../ll/timers.h"      // _Init() _Start() _Stop() _Set_Params()
 #include "lualink.h"           // L_handle_metro()
 #include "io.h"                // IO_handle_timer
-#include "events.h"
 
 typedef enum { METRO_STATUS_RUNNING
              , METRO_STATUS_STOPPED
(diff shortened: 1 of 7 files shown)


lib/metro.c
```diff
@@ -85,14 +84,9 @@ static void Metro_bang( int ix )
 {
     // TODO confirm lua(1) makes a single tick
     if( ix < 2 ){
-        IO_handle_timer( ix );
+        L_queue_in_stream( ix );
```
tehn: confused why this is happening here? how does metro relate to in_stream?

tehn:
looks good except for the one confusing bit in metro.c which i left a comment on.

just to confirm, you tested all changed code on hardware?

trentgill:
Testing on hardware. 'stream' and 'change' have been running with the 33-133Hz ASLs for 5 hours no crash.

The stream handler in the metro library is weird and i've struggled with the best way to approach that. Reality is the metro library is a thin wrapper over the hardware timers. Those timers are setup so the IRQ calls the `Metro_bang()` function in metro.c. The input 'stream's use 2 of those timers dedicated to each input channel. Then I'm just special casing the first 2 and forwarding them for the stream callback. This should be refactored so the stream and metro send different fnptrs to the timer library so that the timer callbacks forward to the correct library.

I've made issue #114 to refactor this code, but would suggest merging this PR as is, because it works and solves the crashing bug by slope callbacks being in the event loop.

PR #119, opened 2019-05-21
update output lib & readme for voltage get/set

trentgill:
The proposed syntax for setting an output channel directly has been updated from `.value` to `.volts` and is implemented and tested. README updated with syntax details. `output[n].volts(3.0)`

Additionally one can read the current output voltage with `volts-right-now = output[n].volts`

And doing a table call like `output[n]()` is equivalent to `output[n].asl:action()` for terseness & it makes sense to me as saying "DO IT" to the output channel.

Updated readme w some details of ASL.
Updated readme to document existing Input mode setter syntax. This feature could be removed i think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,10 @@ input[2].mode( 'stream', 0.05 ) -- sends 20 values per second
 input[1].mode( 'change' ) -- default gate detector
 ```
 
+There is *also* an assignment style:
+`input[1].mode = 'stream'`
+though it is limited to changing mode (no params) and should probably be removed(?)
+
 ### Standalone Inputs
 
 The only real change when using the inputs on crow itself is you'll need to define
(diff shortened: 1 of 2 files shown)


trentgill:
"ie, why wouldn't a user try: `output[1].volts = 2`"
oh that also works! sorry my PR was non-exhaustive. both assignment-to and call-of the member variable will set it, while only reading-from will get.

```
output[1].volts = 3.0
print(output[1].volts)

=> 3.0
```

my inclination is away from get/set, but for now i've just try to enable as many parallel syntax options so we can experiment w writing scripts in different styles. once we have a set of ready-for-launch scripts, i think we should code-review them all for consistency & make decisions about what kind of access methods we want to promote. additionally that would be the right time to consider any naming, capitalization, and data-structuring issues.

i think the output-table call isn't the best idea either. the named actions seem better to me too, and i made some comments about meta control over the ASLs in the readme.

that shortcut should be possible & i agree it makes sense most likely. makes me think the syntax & structuring of the asl should change. 'action' would be a member of output-table to which we assign an asl `output[1].action = Asl.new(1)`. then 'calling' the asl would control execution of the action, with the possibility of passing a parameter: eg. `output[1]:action('restart')`.

it gets complicated, but i think this is a good direction to head. again i'm just adding options and we can see what feels good in real scripts. still not sure how much more functionality i'll be able to get running in asl by release.

tehn:
ah, ok, rad!

```
output[1].volts = 3.0
print(output[1].volts)

=> 3.0
```

is very elegant!

PR #136, opened 2019-06-15
rename return vals

tehn:
`ret_cv` -> `input`
`out_cv` -> `output`
`crow_id` -> `identity`

suggestions welcome!

`change` has remained. i question it's need vs. just being identical to `input` since your script should be able to smartly behave differently if you've switched the input mode.


trentgill:
i'm on the fence / haven't thought enough about use cases.
fundamental difference between 'input' and 'change' is input returns a float, while 'change' returns a bool (0/1 int really). is it more work for users to write a separate function for each, or have to write a handler that works for both?

if you imagine the user leaving it constant in the context of a given script it's not a big deal, but if the user wants to change the behaviour at runtime it might cause issues.

furthermore, i'm planning to add a number of other input 'modes' after 1.0 which would have less equivocal arguments in the handler (eg a note-detector which would give the user a note-number output when the input entered a new note)

tehn:
cool! looking forward to your ideas.

i'm good with `stream`

unless you want `stream` and `query` to be separate handlers. in which case i'd suggest `input` for `query` to match the output query.

are stream and query significantly different use cases? seems sortof? query is sortof a sample-hold?

tehn:
any traction with this idea?

query -> ^^input
stream -> ^^stream
change -> ^^change

trentgill:
query -> ^^input
stream -> ^^stream
change -> ^^change

i can work with that. unless we want query -> ^^query?
happy either way. can you update the README to reflect these changes too?
i'll take care of adding it to the max object.

trentgill:
i've bounced back and forth on this, but i think what you have makes the most sense:
`query` -> `^^input`
`stream` -> `^^stream`
`change` -> `^^change`
`out_cv` -> `^^output`

i *want* to merge input & stream (they are so similar), but i'd insist upon using 'stream' for the name as this aligns with how the user would define an event handler in the crow-local context. problem is it feels awkward to receive a one-off query through a callback called 'stream'. thus i think it's best to keep them separate.

tehn:
how about we have both query and stream just go to input? I don't feel two callbacks are fully necessary

trentgill:
ok. don't think it's worth stressing over it any more. there's always going to be a gap between local/remote modes, so shouldn't be a problem to say it's `input` on norns, and `stream` on crow.

PR #149, opened 2019-07-12
change_handler now passes state as a bool

### LINKED ISSUE #143: input.change function returns number (should return bool)
would make interaction with held{} more elegant

trentgill:
fixes #143 

likely causes some new bugs in both monome/norns but also the max object

### DIFF
--- a/lua/input.lua
+++ b/lua/input.lua
@@ -107,7 +107,7 @@ setmetatable(Input, Input) -- capture the metamethods
 
 -- callback
 function stream_handler( chan, val ) Input.inputs[chan].stream( val ) end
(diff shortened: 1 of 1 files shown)


tehn:
ha should've tested on hardware first:

```
[string "local _crow = {}..."]:37: bad argument #3 to 'tell' (string expected, got boolean)
crow receive: error: input change
```

trentgill:
aye.. yes sorry. will fix shortly.

tehn:
lualink.c fix i think

trentgill:
hrmmm.. seems like `luaL_checkstring()` will coerce a number to a string, but not a boolean :/ will just convert to a string in the input.change event-handler i think.

Issue #144, opened 2019-07-08
raw table containing held doesn't restart with action(bool)


trentgill:
only runs once. works fine if you wrap in a loop, but obvs that's not what we want :/

PR #151, opened 2019-07-13
Fix bug caused by #149 in input[n].change default function

trentgill:
_c.tell lua function is not able to handle booleans directly. using the and/or lua idiom to convert a bool to a 1/0.

bonus is it means for crow standalone applications you get the nice boolean usage (eg. if _ then, or if not _ then) but norns & max remote application is unchanged.


trentgill:
squash & merging bc this fix allows me to keep testing for #144, but lmk if this doesn't make sense.

PR #152, opened 2019-07-13
ASL will restart if the table has been completed

### LINKED ISSUE #144: raw table containing held doesn't restart with action(bool)

trentgill:
added checks to make sure the ASL would restart if it has completed all the actions before the current method call is received.

a number of new test cases added to tests/asl.lua to cover this behaviour.

testing support for the string 'nil' was added to all the tests to easily call through as if no argument was sent to :action.

fixes #144

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,7 +368,27 @@ end
 
 ## Output Library & ASL
 
-TODO
+Each of crow's 4 outputs use an instance of the Output library to give control over
+the CV values on each channel. The most basic actions are designed to be easy but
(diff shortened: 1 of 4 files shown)


trentgill:
updated the README and restructured it to talk about `.volts` and `.slew` before even touching ASL.

also fixed the behaviour where you can pass an ASL as the value to an output[]() call.

PR #159, opened 2019-07-19
added note() and n2v() ASL library functions

isaacpearl:
Added note() and helper n2v() (note to voltage) functions to asllib, which should allow for easier/more readable melodic sequencing in ASL.

### DIFF
--- a/lua/asllib.lua
+++ b/lua/asllib.lua
@@ -18,9 +18,9 @@ function n2v( n )
 end
 
 function note( noteNum, duration )
-	return{ to( n2v(noteNum), 0 ) 
-		  , to( 'here', duration )
-	   	  }
+	return {to( n2v(noteNum), 0 ) 
+           ,to( 'here', duration ) 
+           }
 end
 
 function lfo( speed, curve, level )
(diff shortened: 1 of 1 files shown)


lua/asllib.lua
```diff
@@ -11,6 +11,16 @@ function negate( v )
...
 end
 
+function n2v( n )
+	if type(n) == 'function' then
+		return function () return n()/12 end
+	else return n/12 end
+end
+
+function note( noteNum, duration )
+	return{to(n2v(noteNum), 0), to('here', duration)}
```
trentgill: could you update the layout here to make it match the rest of asllib calls. something like:
```
return{ to( n2v(noteNum), 0 )
      , to( 'here', duration )
      }
```
i think it makes it clearer that it's a chaining of 2 `to`s.

isaacpearl:
done!

isaacpearl:
this is looking weird in the source code on here (vs my editor settings) - working on it still

isaacpearl:
ok! this is actually done now.

PR #166, opened 2019-09-09
ii add

tehn:
adds initial ansible ii support (extender, kria, mp)

### DIFF
--- a/lua/ii/ansible.lua
+++ b/lua/ii/ansible.lua
@@ -75,13 +75,7 @@ do return
   -- no setter for CMD 10 input
   }
 , getters =
-  { { name = 'cv'
-    , cmd  = 8 + get_offset
-    , args = { 'channel', u8 }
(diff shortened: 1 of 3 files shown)


tehn:
also not sure about the submodule updates. should it be on newest master?

trentgill:
This looks great!
I'm guessing you'll want to switch the `volts` style commands over to the new `s16V` type I added. Note it's only signed for now as the s16 range easily covers +/-10V but perhaps this will cause issues if people type `II.ansible.cv(1,-3)`? I'm guessing Ansible just clamps the values, but might be worthwhile seeing what happens...

Regarding the submodules, I think you need to `git pull` in the dfu-stm32f7. It looks like you're on the right branch (`f722`), but don't have the changes when i added dac support to the bootloader (so the outputs no longer sit at 10V while bootloading!)

PR #174, opened 2019-09-13
Cal and Midi become cal and midi

trentgill:
lowercased both `cal` and `midi` so that all user calls to the libraries are through lowercase names.

also removed the `metro = Metro` alias because we only ever use `metro` and that seems correct.

tested now & looks good. also updated README entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ input[1].change = function(state)
 end
 ```
 
-## Midi Library
+## MIDI Library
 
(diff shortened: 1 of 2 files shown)


tehn:
@trentgill what does this need to have added prior to merge?

trentgill:
@tehn nothing afaik. i just wanted to bring it to your attention in case the norns integration had specific support for the `cal` or `midi` commands using the uppercase.

i think this is good to merge.

PR #178, opened 2019-09-18
kill command added to C layer parser

trentgill:
can now use `^^k` to hard-reset the lua environment. particularly useful for developing a standalone application for use with crow when you trash the lua env, but don't want to reset the device (because you haven't implemented auto-reconnect).


tehn:
doc update for this?

tehn:
ah i see it. will get it moved over the monome.org/docs

PR #187, opened 2019-09-26
new default.lua script

trentgill:
plays a dualing / phasing melody. the sequence (rhythm & pitch) are determined by each crow's unique id number, so should be subtly different across all modules.

input[1]: attach trigger to advance sequence
input[2]: near 0V is least movement, +ve & -ve stretch the melodies out. counter-clockwise is ionian, clockwise is pentatonic. there are 2 sequences being crossfaded as the voltage increases
output[1+2]: cv/env for voice1
output[3+4]: cv/env for voice2

note: added extra hooks to get the unique-id into lua.

### DIFF
--- a/lua/crowlib.lua
+++ b/lua/crowlib.lua
@@ -29,7 +29,7 @@ function _crow.libs( lib )
         metro  = dofile('lua/metro.lua')
         ii     = dofile('lua/ii.lua')
         cal    = dofile('lua/calibrate.lua')
-        midi   = dofile('lua/midi.lua')
+        --midi   = dofile('lua/midi.lua')
     elseif type(lib) == 'table' then
         -- load the list 
     else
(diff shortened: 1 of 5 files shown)


lua/crowlib.lua
```diff
@@ -29,7 +29,7 @@ function _crow.libs( lib )
         metro  = dofile('lua/metro.lua')
         ii     = dofile('lua/ii.lua')
         cal    = dofile('lua/calibrate.lua')
-        midi   = dofile('lua/midi.lua')
+        --midi   = dofile('lua/midi.lua')
```
tehn: we can call this manually in our own scripts right? :)

lua/default.lua
```diff
@@ -1,25 +1,97 @@
...
+function env(count)
+  for i=1,2 do t[i] = t[i] + 1 end
+end
+
 function init()
-    -- start all the lfos
-    for c=1, 4 do
-        output[c].asl:action()
+  -- generate unique tables
+  lcg(unique_id())
```
tehn: rad

PR #189, opened 2019-09-26
lfo params updated

trentgill:
lfo now takes a 'time' and divides it by 2 at runtime for accurate timing

@dndrks does this break anything in the m4l stuff?

### DIFF
--- a/lua/asllib.lua
+++ b/lua/asllib.lua
@@ -4,7 +4,12 @@
 --
 Asllib = {} -- how should we refer to the below?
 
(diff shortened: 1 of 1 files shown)


dndrks:
@trentgill , only a little. I'll adjust the m4l device, no worries! I was doing this the patcher, so it saves some code!

trentgill:
no sweat, i'll wait to merge it til you've done the update. sorry for the last minute change!

dndrks:
eh, I'd like to test while I update -- merge away!

PR #190, opened 2019-09-27
add dockerfile + instructions for cross platform firmware development

csboling:
Adds a Dockerfile which sets up all dependencies necessary to compile the firmware on any platform with Docker installed.

The easiest way to get this to work with off-the-shelf Ubuntu packages was to change the Makefile compiler invocation to `arm-none-eabi-gcc` instead of `arm-none-eabi-gcc-4.9.3`. If 4.9.3 is required / desired I can revise this to get a GCC 4.9 environment set up in the Dockerfile instead.


trentgill:
I'm happy for the Makefile to remove that explicit requirement. I think it was put in because I installed gcc 6.x which borked the build (on my mac, but not linux). Only way to know for sure would be to test. 

@csboling does this successfully build?

csboling:
Yeah, it produces a crow.bin, reckon I won't know for sure if it works til some time next week.

tehn:
tested/works!

PR #202, opened 2019-10-07
Correct initialization of lua env for different upload / boot processes

### LINKED ISSUE #33: reset lua environment after new user script received
currently, the script is just loaded, and would require a hard reset to stop existing code.

### LINKED ISSUE #201: Blank slate, Flash upload & initialization procedure
@tehn @ngwese I want to straighten this question out before we merge #193.

### Prop 1: `Lua_Reset()` should *only* restart the lua VM

Currently this function calls `init()` after restarting the VM, which can cause a CPU & RAM usage spike. This is problematic as it happens right as we start blasting the USB port with the script.

If we do this, `^^k` (which just calls `Lua_Reset()`) won't call `init()` & becomes a quick, non-flash-touching way to get the bare minimum lua env with maximum free memory. If the user wants to start their userscript, they can type `init()` in druid.

Reduces RAM footprint while we're trying to malloc the reception buffer for `^^s`.

### Prop 2: `^^f`(or `^^w`) & `^^e` both call `init()` after a successful transmission.

Currently these don't start the userscript, unless the user calls `^^k` or types `init()` or far worse, adds `init()` to the end of their userscript. The problem with this last point (which i've been doing constantly) is it breaks the above `^^k` 'give me a blank slate' idea, and can result in duplicate metro issues.

///////////////

In `druid` this means that `r` is just a faster version of `u` that doesn't wear out the flash memory.

What is not currently possible (in druid), is sending multi-line blocks of text. There's no way to enter it from druid's command line, but also no way of getting a code-chunk executed on crow in a way that doesn't trash the VM.

Basically we need a method that is like the old `r` which can either upload a file 'on top' of the existing VM state, or perhaps something fancier like uploading a named object (eg function, table, variable). The latter would be more complicated, but allow a richer live-coding experience where you can have multiple chunks of code, and selectively upload them.

Obviously this is way off topic for this issue, but it gives some context around usage of the above i think.

trentgill:
- default.lua no longer calls init() at the end of the file (crow itself should do this)
- if a userscript doesn't include an init() fn, lua doesn't have an eval error
- using `^^w` to end an upload & 'write' to flash (leaves `^^f/F` free for First)
- any upload (regardless of write-to-flash or run) calls init() after a successful upload.

fixes #201 
fixes #33

### DIFF
--- a/lib/caw.c
+++ b/lib/caw.c
@@ -62,7 +62,7 @@ static C_cmd_t _find_cmd( char* str, uint32_t len )
                 if(      *pStr == 'b' ){ return C_boot; }
                 else if( *pStr == 's' ){ return C_startupload; }
                 else if( *pStr == 'e' ){ return C_endupload; }
-                else if( *pStr == 'f' ){ return C_flashupload; }
+                else if( *pStr == 'w' ){ return C_flashupload; }
                 else if( *pStr == 'c' ){ return C_flashclear; }
                 else if( *pStr == 'r' ){ return C_restart; }
                 else if( *pStr == 'p' ){ return C_print; }
--- a/lib/lualink.c
+++ b/lib/lualink.c
@@ -81,7 +81,6 @@ void Lua_Reset( void )
 {
     Lua_DeInit();
     Lua_Init();
-    Lua_crowbegin();
 }
 
 void Lua_load_default_script( void )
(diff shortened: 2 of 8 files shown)


tehn:
code looks good. haven't tested yet.

didn't see any mention of `^^f` in the code?

Issue #195, opened 2019-10-03
add bounds checks to asllib functions


trentgill:
eg: `lfo(0)` causes lua to crash with an infinite loop. should clamp time to some minimal values.

currently `lfo` can take a function as an argument, so it should probably have the `if type == 'function' then clamp( arg() ) else clamp( arg ) end` to support that syntax. <- perhaps this suggests a new kind of abstraction as we're making more of these functions that take functions or values.

trentgill:
closed by #226

PR #206, opened 2019-10-07
adds bounds check to lfo

trentgill:
Begins to solve #195 but should also add similar things throughout i guess?
This is just a quick fix to stop hard crashes when time <= 0

### DIFF
--- a/lua/asllib.lua
+++ b/lua/asllib.lua
@@ -8,7 +8,7 @@ Asllib = {} -- how should we refer to the below?
 function boundgz( n )
     if type(n) == 'function' then
         return function()
-            nn = n()
+            local nn = n()
             return (nn <= 0.01) and 0.01 or nn
         end
     else return (n <= 0.01) and 0.01 or n end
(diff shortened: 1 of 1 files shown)


lua/asllib.lua
```diff
@@ -4,6 +4,16 @@
 --
 Asllib = {} -- how should we refer to the below?
 
+-- clamp to bounds, must be greater than zero
+function boundgz( n )
+    if type(n) == 'function' then
+        return function()
+            nn = n()
```
pq: maybe make the `nn` declaration `local`?

lua/asllib.lua
```diff
@@ -4,6 +4,16 @@
...
+            return (nn <= 0.01) and 0.01 or nn
```
pq: curious how you settled on `0.01` to clamp to?  might be useful to doc this somewhere?

lua/asllib.lua
trentgill: good catch!

PR #216, opened 2019-10-12
increase First gate threshold to 1V

csboling:
Seems from measurement that Maths EOR/EOC gate outputs, and probably some other common modules, have a sloppy low of about 0.7V, causing First not to trigger. Not sure if maybe the default threshold (0.5V) in input.lua should be changed as well. This was giving at least [one person](https://llllllll.co/t/crow-help-general-connectivity-device-qs-ecosystem/25866/128) trouble, thought I saw another post about this too but am not finding it now.

### DIFF
--- a/lua/First.lua
+++ b/lua/First.lua
@@ -89,7 +89,10 @@ function init()
   output[4].action = ar(get_a,get_d)
 
   -- start sequence!
-  input[1]{ mode = 'change', direction = 'rising' }
+  input[1]{ mode = 'change'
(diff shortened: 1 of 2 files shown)


tehn:
thanks for spotting this!

trentgill:
Let's not change First.lua, but change the default in input.lua. I want users to be able to use the default parameters as much as possible, so feels better to change the defaults to more widely acceptable values.

trentgill:
Looks great, thanks!

PR #218, opened 2019-10-13
strip a bunch of stale docs, WIP

trentgill:
A lot of stale things have been thrown out. I pulled all the docs I knew for sure had been covered elsewhere. I left a bunch of explanations around inputs & outputs & asl as I wasn't entirely sure they'd been fully captured elsewhere.

Also updated the readme-development.md to clarify that fennel & luarocks are no longer needed for the build process.

Need to spend some more time figuring out exactly what is redundant, and what needs to be copied into the references / scripting tute etc.
CC @dndrks @tehn


tehn:
great! needs a quick conflict resolution then let's merge

PR #227, opened 2019-10-18
Boot messages & inferred script name printing

### LINKED ISSUE #217: Refine prints when crow is booted / restarted
Currently when crow restarts there are a bunch of prints as each library is successfully loaded in the Lua environment. These are pretty much always the same & I've never seen them fail apart from a bogus firmware. Consider that if anything they should be 'debug' prints that aren't in the release version of the firmware.

Instead, does it make sense to print something when crow is booted, or after `^^k`, or after a new program is run/uploaded with `^^s` and `^^w/e`? This could be similar to `^^p` where it can say:
1. 'First' running
2. No user script running
3. User script running

If we wanted to get fancy, the user script could print the first line of the uploaded file (if it's a comment with three dashes) as this is likely the name of the script.

trentgill:
When running `^^c` and `^^F` the successful change is confirmed to the user.

When running / uploading with `druid`, adds confirmation of a successful flash (if ending with `^^w`). Then print `Running: <script name>` where `<script name>` is the first line of the uploaded script, after the comments & any spaces. We can't enforce this, but I think when people see the name of their script appear, they'll want that (and i'll add it to the scripting docs).

Fixes #217.

### DIFF
--- a/lib/flash.c
+++ b/lib/flash.c
@@ -1,5 +1,7 @@
 #include "flash.h"
 
+#include <string.h>
+
 #include "../ll/debug_usart.h"
 
 #define USER_MAGIC 0xA  // bit pattern
@@ -20,10 +22,6 @@ USERSCRIPT_t Flash_which_user_script( void )
         default:         return USERSCRIPT_Default;
     }
 }
-//uint8_t Flash_is_user_script( void )
-//{
-//    return (USER_MAGIC == (0xF & (*(__IO uint32_t*)USER_SCRIPT_LOCATION)));
-//}
 
 void Flash_default_user_script( void )
 {
@@ -123,6 +121,31 @@ uint8_t Flash_read_user_script( char* buffer )
         sd_addr += 4;
     }
     return 0;
+}
+
+char* Flash_script_name( void )
+{
+    static char script[32];
+    memset( script, '\0', 32 );
+    switch( Flash_which_user_script() ){
+        case USERSCRIPT_Clear: strcpy( script, "No user script running." ); break;
+        case USERSCRIPT_Default: strcpy( script, "'First' script running." ); break;
+        case USERSCRIPT_User:
+            strcpy( script, "Running: " );
+            char* name = (char*)(USER_SCRIPT_LOCATION + 4);
+            while( *name == '-' ){ name++; } // skip commments
+            while( *name == ' ' ){ name++; } // skip spaces
+            char* linebreak = strchr( name, '\n' );
+            if( linebreak ){ // print until newline
+                int len = (int)(linebreak - name);
+                if( len > 22 ){ len = 22; }
+                strncpy( &script[9], name, len );
+            } else { // can't find a new line, so just fill the buffer
+                strncpy( &script[9], name, 22 );
+            }
+            break;
+    }
+    return script;
 }
 
 // CALIBRATION //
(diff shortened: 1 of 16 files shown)


lib/flash.c
```diff
@@ -125,6 +123,39 @@ uint8_t Flash_read_user_script( char* buffer )
...
+    }
+    return dest;
+}
+
+char* Flash_script_name( void )
+{
+    static char script[32];
+    memset( script, '\0', 32 );
+    switch( Flash_which_user_script() ){
+        case USERSCRIPT_Clear: strcpy( script, "User script cleared. Running." ); break;
```
simonvanderveldt: This is the case when there is no script?
If so wouldn't something like "No user script" be a clearer message? The "Running" at the end here is throwing me off, what is it running if there's no script?

And it seems these messages duplicate the ones here https://github.com/monome/crow/pull/227/files#diff-4c329923d1fdc04b8b83730be099f7dfR119
Could we merge them/the logic somehow?

trentgill: I agree it'd be nice to merge the logic.

Originally I had the same "No user script" message, but it's strange because it's what you see immediately after typing `^^c`, so it feels really strange. To me, responding to a user command with "No user script" *feels* like an error message, rather than confirmation of what the user just performed. I agree re: 'running'. will remove.

At present, the printing on boot isn't even working for some reason (an existing issue i had forgotten), so for now they are better described as the messages you see after changing what the user script is.

trentgill:
On restart, unfortunately the prints aren't working for some reason. Only prints from Lua were coming out before the init() function was called which I can't (right now) explain.

I opted for the "use the top-line comment" because 1) it's way easier to implement, 2) doesn't break druid, 3) is familiar to norns scripters as that's how the on-board script-docs are generated.

Perhaps i'm being lazy, but it feels elegant & logical that the first line in your script would be the name of the script.

lib/flash.c
simonvanderveldt: Re: `^^c` I guess the only place/moment we can confirm to the user that the script has been cleared is immediately after clearing it (I guess that would be in `Flash_clear_user_script`).

I guess we could even omit passing the script name to `Lua_crowbegin` in that case?

trentgill: Updated the PR to consolidate the logic for 'print' and 'which script'.

trentgill:
Restructured a few things to cleanup main.c and flash.c. all the flash-touching commands happen via the repl library as they are alt versions of the 'upload' procedure.

lib/repl.c
```diff
@@ -76,21 +77,44 @@ void REPL_upload( int flash )
...
+    Caw_send_luachunk("User script cleared.");
+    REPL_print_script_name(NULL);
+    Lua_crowbegin();
+}
+
+void REPL_default_script( void )
+{
+    Lua_Reset();
+    Flash_default_user_script();
+    Caw_send_luachunk("Using default script.");
```
simonvanderveldt: Seems like we're using both "default" and "First" for this. Should we just use "First" here for clarity?

tehn:
looks good

lib/repl.c
trentgill: I'm going to leave it because it is *always* followed by `"Running: First.lua"`. Folks seem to have a hard time with whether 'First' is getting replaced when they upload a script, so calling it the `default` script I think helps with seeing it as 3rd category.

PR #234, opened 2019-10-21
Stop all interrupt generating routines when calling Lua_Reset()

trentgill:
This is an alternative PR for @csboling 's #232. Firstly, big thanks for isolating this issue!!

This PR takes the idea of turning off the input streams when uploading, and applies it to all C routines that generate interrupts causing events to be added to the queue. My hypothesis is that the crashes are resulting from events being added to the event queue after the Lua VM has been torn down, then being called again.

This is acheived in the C-function `Lua_Reset` which is called before either a `r` or `u` upload in druid. As we're tearing down the whole Lua VM, the metros should not be restarted manually, instead waiting for the lua script to do it if that's part of the init() function. Now Lua_Reset does the following before calling Lua_DeInit():
- deactivate inputs in either `stream` or `change` mode
- ensure all outputs are set to zero volts & don't raise events when they arrive there
- deactivate all metros
- empty the event queue

Testing here shows I can repeatedly `u` a long script on linux. I'll test on OS X & Windows (unless someone beats me to it).

### DIFF
--- a/lib/events.c
+++ b/lib/events.c
@@ -25,17 +25,8 @@ void (*app_event_handlers[E_COUNT])(event_t *e) = { handler_none };
 // initialize event handler
 void events_init() {
     printf("\ninitializing event handler\n");
-    int k;
 
-    // set queue (circular list) to empty
-    putIdx = 0;
-    getIdx = 0;
-
-    // zero out the event records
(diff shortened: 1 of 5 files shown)


simonvanderveldt:
This might fix https://github.com/monome/druid/issues/44

PR #244, opened 2019-10-24
avoid undefined behavior in REPL_print_script_name

csboling:
This patch uses a scope to get around the compiler error preventing variable declarations as the first line of a `case`, instead of the existing `buffer[0] = buffer[0];` statement. In some cases (e.g. after script upload) the param`buffer` is null, but compilers are allowed to assume that any pointer that's dereferenced is not null. Presumably GCC 5.4.0 (in my docker environment) does assume this, and uses it to inline the ternary expression conditional on `buffer`, whereas GCC 4.9.3 behaves differently. This causes a null pointer to be passed to `REPL_script_name_from_mem`, resulting in [this](https://github.com/monome/druid/pull/57#issuecomment-545708063) mysterious 54 bytes of uninitialized junk in the USB buffer followed by a UTF8 decoding error in druid. This patch resolves that error and I believe makes https://github.com/monome/druid/pull/57 unnecessary.

### DIFF
--- a/lib/repl.c
+++ b/lib/repl.c
@@ -158,8 +158,7 @@ void REPL_print_script_name( char* buffer )
         case USERSCRIPT_Default:
             Caw_send_luachunk("Running: First.lua");
             break;
-        case USERSCRIPT_User:
-            buffer[0] = buffer[0]; // satisfy switch
+        case USERSCRIPT_User: {
             char script[64];
             memset( script, '\0', 64 );
             strcpy( script, "Running: " );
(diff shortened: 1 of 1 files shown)


trentgill:
Woah amazing catch! I never understood why switch statements complained like that, and this has been my quick hack to try things out that i often forget to fix. Didn't know that adding a scope was sufficient to make switch happy.

There's another piece of code I'm working on that will benefit from this, so double thanks!

PR #248, opened 2019-11-02
I2C LeaderTx queue & LeaderRx returns via event system

### LINKED ISSUE #246: ii 'get' requests shouldn't call lua directly
currently the L_handle_ii_leadRx() function is called directly from the i2c callback. This can cause the Lua VM to crash if it's currently doing something else when the callback occurs.

Instead, the returned data should be posted in the event queue to be processed at the next possible time.

trentgill:
Overhauled the i2c driver & application layer for Leader operations.

When sending commands to other modules there is now a queue (length=16) so using a sequence of commands in an init script will work through the list (asynchronously) until all have been processed. The queue is built into the `lib/ii` files rather than the low-level i2c driver.

When requesting values from other modules, the I2C driver callback on reception was calling directly into the Lua environment. This PR adds a new event type (and modifies handling of the event.index) such that the received value is processed by the event system. 

Fixes #246 

@whimsical-sam @dndrks this PR should denecessitate the message throttling you're using in the ^^jf m4l patch (and in the max object?)


tehn:
this is super smart, thanks

lib/events.c
```diff
@@ -95,8 +96,7 @@ uint8_t event_post( event_t *e ) {
         }
     );
 
-    //if (!status)
-    //  printf("\r\n event queue full!");
+    if( !status ){ printf("event queue full!\n"); }
```
tehn: does it make sense to leave these debug prints?

trentgill:
I think it makes sense for it to persist in the dev build (printf only sends to the uart). I’m assuming it gets optimized out with `make R=1` but i should check that assumption...

Issue #221, opened 2019-10-14
Many ii ansible commands overwhelm or partially lockup ansible 


ngwese:
There are potentially multiple factors at play here. The short version is that calling `ii.ansible.*` functions in rapid succession (such as within a loop) results in commands being partially dropped/ignored at best or ansible partially/fully locking up at worst.

Test setup:
- crow with 1.0.0 firmware; ansible w/ 2.0.0 firmware
- crow ii connected directly to ansible ii (no other devices on the bus

To reproduce:
- power up devices; connect to crow via druid and run:
- `ii.pullup(true)`
- `for i=0,3 do ii.ansible.cv_set(i, (i+1)*2) end`
- `for i=0,3 do ii.ansible.trigger_toggle(i) end`

When setting the cv values in a loop typically only the last cv output appears to get set.

When toggling the tr outputs via a loop run from druid the resulting tr state is somewhat random.

In at least one case ansible stopped responding.

trentgill:
This is likely because the i2c library isn't queuing commands.
Surprised this is causing ansible to freeze though. Is there a script that makes ansible freeze reliably?

ngwese:
I don't have a script which reproduces the ansible freeze. It happened while running the above loops interactively while testing. I'm not 100% certain that ansible was completely frozen but it definitely stopped responding to `ii` messages (until I power cycled it).

csboling:
Haven't seen a crash yet in multiple runs of e.g. 
```lua
for k=1,1000 do for j=1,1000 do for i=0,3 do ii.ansible.trigger_toggle(i) end end end
```
with release firmwares. I do have several more modules on the bus, in particular a Teletype (black PCB) is providing pullups. Teletype can also be sending I2C periodically while crow is spamming the bus like this, including TT -> Ansible messages as well as messages to other modules, and nothing has crashed. Will try again later with just crow pullups maybe? It is also possible that this PR https://github.com/monome/ansible/pull/72 to get this libavr32 patch https://github.com/monome/ansible/pull/72 could result in fewer I2C-related Ansible crashes.

trentgill:
@ngwese I know you didn't have a clear crash case, but wondering if this problem has been solved with 1.0.2?

ngwese:
I was never able to find a consistent repo case but I could repo it.

I’ll experiment with 1.0.2 this weekend if possible.

Issue #55, opened 2019-02-14
teletype support for crow


trentgill:
crow is able to be queried by another i2c leader device. teletype is the clear companion device who would lead crow, so the TT project must be updated with hooks for the actions.

This should be implemented for a single crow device first.

Afterwards we can consider how to handle multiple crow devices. This could be sequential input/output numbers (ie `CROW.OUT 5` would refer to the first output of the second device, vs `CROW.2.OUT 1` or `CROW.OUT 2 1` etc).

Proposed initial functions are listed in lua form: https://github.com/trentgill/crow/blob/aa45d96a322f6c818e84a98bc2d69b58a189ed45/lua/ii.lua#L38
and TT form: https://github.com/trentgill/crow/blob/aa45d96a322f6c818e84a98bc2d69b58a189ed45/lua/ii.lua#L67

All the generic functions should send /receive int16 values.

trentgill:
to aid integration between crow & TT, a small library of helper functions should be added as a lua file.

purpose is to allow TT familiar users to have a way to treat values in crow the same as they do in TT without worrying about the transformations they hide:
`TT.V(x)` converts a lua volt (ie float where 1.0 is 1V) to a TT volt (ie 1638.4)
`TT.VV(x)` as above but for 163.84
`TT.N(x)` as above but 136.53

note the i2c driver will truncate floating point values, so this lib might need to add 0.5 after confirming the rounding behaviour or TT.

this will be deprecated after #49 is complete.

scanner-darkly:
i can help with the teletype side of things once the supported messages are finalized.

trentgill:
included in the TT 4.0 beta

PR #250, opened 2019-11-08
Crow as II-follower

### LINKED ISSUE #17: ii follower commands for crow
https://github.com/trentgill/crow/blob/b2d5f22be6f6639a3cf1eaf6bebc0865c440a421/lua/ii.lua#L36  proposes a set of crow ii commands (as follower). it provides the basics of input / output, in as similar a way as possible to teletype (so it can be used as an expander with almost zero mental overhead). 

then also the generic `.CMD` functions create a generic interface to get data into/out of crow. the implementation is entirely up to the user. all that is predetermined is the number of args (with options up to 4 args), which the user can then deal with in their crow script. eg. `CW.CMD2` could be used to execute `set_lfo( channel, speed )`

`CW.CMD2` could alternatively be used in `dispatch_ii( action, argument )`, meaning that the first arg actually chooses between a (potentially giant) list of actions.

obviously this allows the user to create utterly undecipherable TT scripts. this seems like a downside, but it also seems like a massive boon for the creative possibilities. the idea that TT can trigger 'some event' with an ii call to crow seems essential, and i'd prefer not to have to codify all the possibilities of what that event should be.

//

an aside: moving to a multi-leader i2c setup suggests a set of new TT decision questions not yet engaged with. specifically, "what functionality should TT provide as an i2c follower?". the immediate parallel to the above proposition is to allow calls to IN, OUT but also SCRIPT.

food for thought!

### LINKED ISSUE #220: II jf control has some stability quirks on startup.
I was having some issues with a script that I’m working on that controls jf via i2c.  It sometimes doesn’t start correctly.  I haven’t quite nailed down the issue, but I believe it is some sort of race condition on load.  Some thoughts:

- jf sent the ii enable message before the system has finished booting up
- a play_note command runs before the ii enablement happens.

When I set my script to run on system power on, it’s like I will hear a handful of the notes, and then it will stop behaving correctly (something like 1n, 2n, 5n, 6n leds will be constantly lit)

I’ve restructured the code of my script to delay i2c enablement and guarded the play note command until it has been enabled, and I believe this has fixed stability.

Will share the script tomorrow, just wanted to get the issue up while I had a second.

trentgill:
crow now successfully receives i2c commands and queries from other i2c leader devices. Has been tested only with two crows talking to each other, either by switching leader/follower roles dynamically, or by one device leading another & querying values.

Current usage (to be added to docs):
```lua
--- leader crow talks to another crow
-- same as other i2c devices. see the list with: ii.crow.help()
function tell_crow_to_do_things()
  ii.crow.output(1,3.2) -- set output 1 to 3.2 volts
  ii.crow.slew(2,100) -- set output 2 to a slew time of 100ms
    -- using ms as an integer to work with Teletype natively :/
  ii.crow.call1(3) -- generic call to crow (signed 16bit integers)
  -- call2 and call3 supported, up to...
  ii.crow.call4(4,3,2,1) -- note floats are *not* supported to make this teletype native
end

function ask_crow_for_info()
  ii.crow.get('input',2) -- get voltage at input[2]
  ii.crow.get('output',3) -- get voltage at output[3]
  ii.crow.get('query0') -- send a generic query with no args
  -- query1 & 2 up to..
  ii.crow.get('query3',5,4,3) -- query crow's response to the provided args
end

-- all getters return a single value in the 'data' arg below
ii.crow.event = function( e, data )
  if e == 'input' then -- the voltage of a remote input (float)
  elseif e == 'output' then -- the voltage of a remote output (float)
  elseif e == 'query0' then -- the response to a zero-arg query (16bit integer)
  elseif e == 'query3' then -- the response to a 3-arg query (16bit integer)
  end
end
```

When implementing crow's i2c follower behaviour:
```lua
-- calls to `output` and `slew` automatically set their respective params (eg output[1].volts = _)
ii._c.call1 = function(a)
  -- do something with the received value 'a'
end

ii._c.call4 = function(a,b,c,d)
  -- do something with up to 4 args
  -- eg: 'a' could be the index into a table of functions that take up to 3 args
end

-- for queries, 'input' and 'output' are handled by the C driver, so lua only sees queryN requests:
-- NOTE: these functions are called directly from the driver & *must* return their response
-- they will block the i2c bus until they return (which will lockup a TT request, but not a crow)
ii._c.query0 = function()
  return unique_id() -- here we use the 'query0' command to return this crow's unique id
end

ii._c.query3 = function(a,b,c)
  -- here we just show that we have access to the lua env:
  return a+b*c
end
```
First question is whether there is a better name for 'this crow' than '_c' (which is a shortcut to the `crow` table i was using elsewhere, but has largely been deprecated). Suggestions welcome. We could change the follower commands to `ii.crow` and leaders to something different `ii.remotecrow`. or use `ii.this` or `ii.self` for the OO crew.

Second question is whether the `queryN` callbacks should share a single callback like the `ii.<device>.get()` commands as they are conceptually similar. Key differences are that queryN returns `N` args rather than just one, so we'd need to use varargs (`...`). Secondly queryN functions *must* return the i2c response from that function call. If no return value is provided they will return zero.

I note there is a danger here that `query` messages could crash crow if the lua environment is currently active when the i2c interrupt occurs. To avoid this, we'd need to overhaul the event system to have a priority stack which feels like overkill until we start seeing crashes (i haven't seen any). As crow is a leader/follower you can send a `callN` message *to* a crow which knows to respond with a matching `callN` which is fine for crow<->crow, but not Teletype (as it can't currently act as a follower). This issue does not apply to 'input' and 'output' queries as they fetch responses in the C layer.

Adding Teletype support (#55) should now be straightforward for anyone who wants to jump on that, though decisions about naming here should be reflected there.

I2C chaining (#5) should now be relatively simple to support by adding some crowlib commands to set reception i2c address. *Note that sending arbitrary Lua is not yet possible as the i2c lib only buffers messages up to 9bytes (1 cmd byte, and 8 arg bytes).*

Fixes #17 
Fixes #220 
Improves (but doesn't necessarily close) #221 (up to 16 queued commands won't be dropped)

Here's an example script where 2 crows work together to count as fast as possible. It will make `druid` eat a lot of CPU cycles as it processes & displays the data.
```lua
--- I2C Handshake

function init()
    ii.crow.get('input')
end

ii.crow.event = function(e,n)
    if e == 'input' then
        if n > input[1]() then
            countup()
        end
    end
end

function countup()
    ii.crow.call1(0)
end

ii._c.call1 = function(a)
    print(a)
    ii.crow.call1(a+1)
end
```


simonvanderveldt:
> First question is whether there is a better name for 'this crow' than '_c' (which is a shortcut to the crow table i was using elsewhere, but has largely been deprecated). Suggestions welcome. We could change the follower commands to ii.crow and leaders to something different ii.remotecrow. or use ii.this or ii.self for the OO crew.

`ii.this` or `ii.self` makes the most sense to me.


Possibly stupid question but the `callN` functions and and `queryN` arguments are fixed? Or can anyone create whatever they want?

And is there a reason we can't use something like `ii.crow.input[n].volts` and `ii.crow.output[n].volts` to mimic the regular queries instead of using `.get()`?

tehn:
super rad, nice work!

opinions:

1. `self`
2. single callback

i'd like to give this a prelim test phase before doing the TT side, to make sure we like the syntax.

tehn (approved):

trentgill:
`ii.self` sounds good.

@simonvanderveldt 
The `callN` and `queryN` functions are limited by the i2c driver-interface. We have to buffer the values so that multiple calls don't block execution waiting for the i2c bus which means a finite cap on message length & thus argument count. 4 args seems like overkill already.

We could collapse them into a single `ii.crow.call()` and `ii.crow.query()` but that would require some magic under the hood. My reason for explicit numbering is that the user could explicitly use the functions for different classes of action. I just tried writing some examples and they didn't make sense immediately, so perhaps example scripts are needed before we make this decision?

@tehn 
re: Single callback: I'm unsure what the most fluid version of this would be. Thoughts / suggestions?
```lua
-- most programmer-y
ii.self.query = function(...)
    local args = {...}
    local nargs = #args
    -- user can switch on nargs, or just handle them all the same?
    return _
end

-- basic helper for count-of-args so it's obvious how to switch based on 'query0' vs 'query2'
ii.self.query = function(count,...)
    local args = {...}
    -- switch on count
    return _
end
-- args are explicit to avoid varargs (which are weird in that you have to explicitly table-ify them
ii.self.query = function(count,a,b,c)
    -- switch on count
    -- a,b, and/or c may be nil
    return _
end

-- provide args wrapped in a table with a count
ii.self.query = function(count,arg-table)
    -- switch on count
    -- use args[n] where n=1:3 (args[n] can be nil)
    return _
end
```

trentgill:
@simonvanderveldt @tehn updated to `ii.self` for the follower callbacks.

simonvanderveldt:
`ii.self` looks good!

> I just tried writing some examples and they didn't make sense immediately, so perhaps example scripts are needed before we make this decision?

The main this I wasn't sure about what it would look like when used in practice/which cases are relevant, so I think some examples are a good idea.

Dewb:
@trentgill: 
> I note there is a danger here that query messages could crash crow if the lua environment is currently active when the i2c interrupt occurs. To avoid this, we'd need to overhaul the event system to have a priority stack which feels like overkill until we start seeing crashes (i haven't seen any).

I found a reproducible crash with query messages: #492 

If I understand the above paragraph correctly, the proposed solution is to have the i2c query handler push an event onto the main event queue to solve the concurrency problem, but then convert the event system to a priority queue so the i2c request can be serviced quickly. Do I have that right? If so, is a queue with arbitrary priorities necessary or would two separate queues work (one for interrupt-priority events, one for normal priority events)?

tehn:
that sounds correct to me.

Issue #68, opened 2019-02-20
absolute shapers for slope library


trentgill:
'absolute' shapers from `readme-development`

trentgill:
```
-- 2 categories of shapers are provided
--      1. relative shapers (like RC curves)
--      2. absolute shapers (quantizers)
--
-- relative shapers should be simple lookup tables with interpolation.
-- these can be relatively small tables as they are not for audio so
-- distortion isn't much of a concern.
--
-- these will cover the basics:
--  linear (which is just pass through)
--  square
--  log / expo
--  sine / cosine / tan
--  sec / csc
--  1/x
--
-- of note here is that all of these relative shapers for drawing a
-- line between 2 points could be applied absolutely as well, shifting
-- the baseline of the output. consider that these could be different!
--
--
-- the absolute shapers are interesting though as a set of divisions
-- thinking primarily in terms of absolutely dividing up the output
-- space.
-- user provides:
--  zero point to base off (could be a 'key')
--  fold point (1V becomes 1 octave, but could be eg 1v2 for buchla)
--  either | divisions (divides fold-size into n equal steps)
--         | list (divides fold-size unequally according a list)
--
-- the natural use is to provide scales for quantizing. by using the
-- list approach, one can send arbitrary pitches, so this could be
-- tonal scales directly, rather than semitones of 'divisions'.
--
-- list can also take fractions to natively handle just intonation,
-- readily accepting a scale as a list of fractions.
--
-- a weighting mechanism could be added to give a wider window to
-- preferred values.
--
-- setting divisions to a large number, then manipulating fold will
-- essentially become a bit crusher with continuously variable bits
--

local shaper = { zero = 0
               , fold = 1  --volts
               , divs = 12 --semitones
               , list = {} --empty means nil
               }

function S_absolute( samp, divs, fold )
    div_sz  = fold / divs
    bracket = math.floor(samp / fold)
    shelf   = samp % fold
    shelf1  = shelf / fold
    div     = math.floor(0.5 + (shelf1 * divs))
    return ( div * div_sz
             + (bracket * fold)
           )
end

function S_absolute( samp, divs, fold )
    local bracket = math.floor(samp / fold)
    local shelf   = (samp % fold) / fold
    local plateau = 0
    if type(divs) == 'table' then
        if divs[1] == 'ji' then
            --TODO needs log shaping i believe
            plateau = divs[2][math.floor(1.5+(shelf * #divs[2]))] - 1
        else
[...]

Issue #249, opened 2019-11-04
Add a 'delay' function to postpone execution


trentgill:
Particularly when using a gate input to sample a voltage input, there can be some lag on the CV side causing pitch droop. This is being noted when using Kria or Teletype to send a 'simultaneous' gate & CV, that the CV has not yet arrived at the destination. There is always going to be some amount of slack in the hardware implementations... so a solution:

The clear way of implementing a 'delay' function is just to use a metro with count=1, so I propose this:
```
function delay( time, action )
  local function devent()
    action()
    metro.free(dmetro.id)
  end
  local dmetro = metro.init( devent, time, 1 )
end
```

This is untested presently.

Thus for a user an existing script is changed like so:
```
-- existing version
input[1].change = function(s)
  ii.jf.play_note( input[2].volts, 5 )
end

-- proposed version, delays the function call by 3ms
input[1].change = function(s)
  delay( 0.003, function() ii.jf.play_note( input[2].volts, 5 ) end )
end

-- proposed version, using a regular named function instead
input[1].change = function(s)
  delay( 0.003, capture_note )
end

function capture_note()
  ii.jf.play_note( input[2].volts, 5 )
end
```

I see a couple of issues worth discussing:
- Using an anonymous function to wrap the delayed body is unfamiliar to many scripters(?)
- Alternatively, using named functions makes this construct far less elegant / terse
- Dynamically assigning metros allows multiple delay timers & named metros, but breaks the ability to use `metro[n]` directly (as the metro.init function doesn't know it is being used)
- In order to `metro.free` the allocated metro the user-fn is wrapped in another function which feels like it could be optimized (how?)
- Queueing up a number of delays could easily use up all the metros (there's only 7 available)

An alternative approach could be to implement this in the C layer using a single hardware timer (reducing lua timers to just 6) and keeping a queue. This would allow many more timers (as many buffers as we're willing to assign, say ~32), and side-steps the issue where `delay` could collide with `metro[n]`. The obvious con of this approach is it's *far* more code. Perhaps this kind of implementation should be considered after first seeing if the lua-implementation is useful / overwhelming the metro system?g

whimsical-sam:
Why not have a queue in Lua?  Something like this -

```
queue = {} -- stores tables with an action and a time

-- add a function to the queue
function schedule(time,action)
    -- schedule event relative to scheduler metro's current count
    local destination = scheduler.count+time*1000 
    local event = {}
    event.time = destination
    event.action = action
[...]

PR #251, opened 2019-11-13
Slope shaping

trentgill:
Update: Complete & ready for testing then merge.

slopes library has been refactored:
- step function is broken into separate pieces according to assumptions about the sample block
- breakpoint blocks are handled recursively per-sample (the vector math was wonky before)
- a shaper function is applied at the end of the step function (can be block or persample)
- each slope is normalized to a [0,1] range and then remapped after shaping

TODO:
- [x] testing. initial reports suggest they sound great as envelopes!
- [x] add vectorized versions of the shapers (everything is done per-sample atm)
- [x] the shapers should be put into a library
- [x] add square shapes ('rise+wait' = `note()` in asllib, 'wait+rise' = `{to('here',t), to(d,0)}`
- [x] update docs for usage
- [x] add waveshape option to `lfo`, `ar` and `adsr` in asllib
- [x] `pulse` in asllib can be simplified with square-rising

completes first half of #68


trentgill:
Usage updates are in the [PR on monome/docs](https://github.com/monome/docs/pull/210). Briefly, all slopes are now able to be *shaped* natively by the slopes C engine.

Options are:
* 'linear': previously the only option. ramp steadily to the destination value
* 'expo': gentle envelopes, and long feeling decays
* 'log': for aggressive or 'plucky' envelopes
* 'sine': nice for LFOs
* 'now': go immediately to the destination then wait for *time* seconds
* 'wait': wait for *time* seconds then go immediately to the destination
* 'over': similar to 'log' but overshoots before landing gently
* 'under': similar to 'expo' but first goes the wrong direction to 'gather momentum'
* 'rebound': simulates a bouncing ball

```lua
-- using the output.volts meta-action
-- eg: ramp to 2V over 200ms, with an exponential path
output[1].slew  = 0.2
output[1].shape = 'expo'
output[1].volts = 2.0

-- additional argument to the to() function
output[1].action = to( 3.0, 0.1, 'over' )

-- updated asllib functions
lfo( time, level, shape ) -- defaults to 'sine'
ar( attack, release, level, shape ) -- defaults to 'log'
adsr( attack, decay, sustain, release, shape ) -- defaults to 'linear'
```

dndrks:
> @dndrks can you help doc this?

looks like trent beat me to it! I'll review the PR once this is live, tho!

tehn:
perfect!

@trentgill this opens up some great new usage for crow, super excited to play it

trentgill:
There's a binary [here](https://github.com/monome/crow/suites/340543728/artifacts/536081) thanks to @simonvanderveldt 's work on the github Actions :) Would love to have some folks try it before we go public, just to make sure the syntax & names are good, but also that the changes to asllib (lfo & ar shapes) are an improvement and not just showing off the capabilities.

simonvanderveldt:
I'd like to help test, but will only have time either Friday or the coming weekend.

dndrks:
@trentgill  -- just uploaded the .bin you posted and getting super wonky results with n2v in max and druid.

**experienced:**
`output[1].volts = n2v(3)` gives a super high note, executing again gives a super low note, high note, low note, etc.

**expected:**
`output[1].volts = n2v(3)` plays a note 3 semitones higher than base note

lua/asllib.lua
```diff
@@ -15,18 +15,19 @@ function clamp( input, min, max )
 end
 
 function note( noteNum, duration )
-    return{ to( n2v(noteNum), 0 )
-          , to( 'here', duration  )
-          }
+    return to( n2v(noteNum), duration, 'now' )
```
dndrks: should this be `return{ to( n2v(noteNum), duration, 'now' ) }` ?

trentgill: This is fine- a standalone `to()` is a valid ASL by itself. You only need the table if you have more than one action. `to` is similar to the other constructs like `loop`, ie they are just functions that return ASLs `loop{ to(), to() }`.

Sorry I still haven't done the full write up of ASL :(

trentgill:
Thanks for catching this bug @dndrks. The changes I made had broken `to()` calls with time=0 which is exactly what `output[n].volts = ...` is doing under the hood (when output.slew == 0). It's all working for me now. You can download a new bin by clicking `show all checks` immediately above this comment, and following through to 'build artifacts'.

trentgill:
@simonvanderveldt 
> I'd like to help test, but will only have time either Friday or the coming weekend.
> 
> [edit] Quick question: How is "wait" different from the delay that was introduced in #249 ?

It's pretty orthogonal. This PR is about making fancy CV curves directly from the ASL syntax (by waveshaping the output samples). The `delay` function is about delaying execution of an arbitrary lua function. These shapes are all applied in the DSP loop (in C), and are only accessible through the list of names above.

For example. You can use the `lfo` asllib function to assign an LFO to an output. If you set the 'shape' to 'now', you get a square-wave that starts with the high-state. Conversely if you use 'wait', you get an LFO that starts in the low-state.

dndrks:
@trentgill , I've tested about 6 hours in the last two days as I revamped the m4l devices to take advantage of new shapes. it all seems rock solid, really really into the expressivity these shapes open up! generally, everything feels really stable -- I only hung once after sending a metric shit ton of bad commands due to a patching mistake :)

trentgill:
Great! Merging this now.

@dndrks re: naming. Why not just have a single canonical string for each? Under the hood, you only need as many chars as to make unambiguous (ie `e` for exponential, `lo` for logarithmic, `li` for linear), but that's obviously unreadable. The performance benefit of shorter names is likely dwarfed by the overhead of the lua call anyway. I think the list I posted in the examples comment above is probably good if accompanied by longer form docs?

PR #252, opened 2019-11-17
JF v3.2 i2c support

trentgill:
Adds crow support for new i2c features in Just Friends v3.2 (forthcoming).

One new setter:
- `voice_pitch(channel,pitch)` is like `play_voice` but doesn't trigger the envelope for legato playing.

Getters:
- `run_mode`: returns the just-type run status (not the hardware jack). 0=off, 1=on
- `mode`: returns the synthesis/geode mode setting: 0=normal mode, 1=synthesis/geode
- `speed`: returns the setting of the speed switch: 0=shape, 1=sound
- `tsc`: returns the setting of the MODE switch: 0=transient, 1=sustain, 2=cycle

Questions:
- should boolean states be 0/1 (not 1/2?)
- should the 'tsc' switch be 1/2/3 (probably it's lua, and it's not a boolean?)
- naming of 'mode' (changing to synthesis/geode mode) is a weird clash with the JF front panel 'MODE' for the 'tsc' switch. Would like to change but would break existing scripts (probably not worth it).


trentgill:
This is ready for final testing. Requires new JF & Crow firmwares. JF3.2 to be co-released with Crow 2.0

Issue #40, opened 2019-01-26
improve 'l2h' lua to c-header converter script


trentgill:
ed note: This whole process could likely be deprecated if some smart person figures out how we can upload lua bytecode to crow directly, rather than manually convert everything into C strings & load them at runtime. The key benefit here is that lua syntax errors will be caught by that process (using `luac` i guess), and perhaps smaller RAM footprint (because the C strings aren't loaded into memory at all).

I'm unclear on the memory implications of the current approach - Some help profiling the memory usage would likely be a good use of time.

/////////////////////////////////////////////////////

_nb: `l2h` is written using 'fennel', which is a lisp syntax for the lua language. if you want to rewrite it in plain lua, that's ok, but you'll need to update the Makefile to stop it from overwriting your work!_

no attempt is made to validate the crow standard libs before 'make' creates binaries. while it's not the role of `l2h` to do full testing of these libraries, it is useful to have l2h validate the code as at least able to be parsed. all that should be necessary is to use loadfile() on the .lua file and return an error if it fails to parse.
- [ ] return with error if lua code has syntax errors

currently the lua files are simply copied, line-by-line, into a c-header file. in particular this means that comments & whitespace are taking up space in flash, and must be parsed out by the lua interpreter at runtime. importantly, once the lua code is flashed onto the chip there is no way to retrieve it, so it doesn't need to stay human readable.

we can reduce flash usage, and RAM overhead at load time the following ways. top of the list is easiest & biggest impact, while end of list is either difficult or has only a small benefit:
- [x] remove comment lines
- [x] remove empty lines
- [x] strip in-line comments
- [x] strip leading whitespace per line
- [ ] don't add a newline character at end of each line unless necessary (is it ever?)
- [ ] remove all other non-functional whitespace
- [ ] rename local arguments with known scope to single-character names
- [ ] precalculate known constants
**note! most of the above is done by the lua interpreter / compiler (ie `luac`), so perhaps a more lucrative use of time is to find a way to upload precompiled lua bytecode, rather than raw lua strings. if taking this approach, we should make sure to compare binary sizes as bytecode can apparently be larger than textual representation.**

[...]

PR #256, opened 2019-11-18
Add CI using GitHub Actions

simonvanderveldt:
Currently simply builds Crow firmware

Basics are there, it builds the firmware, although there's still a failure because of #40 so I don't think the built firmware works. IMHO we should fix that separately.

It will also create a release build when a new release is made.

/cc @csboling

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -1,0 +1,22 @@
+name: Build release firmware
+on:
+  release:
+    types: [created, edited]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+      with:
+        submodules: true
+    - name: Build
+      uses: ./.github/actions/build-firmware
+      with:
+        release: '1'
+    - name: Upload artifact
+      uses: actions/upload-artifact@master
+      with:
+        name: crow-${{ github.sha }}.bin
+        path: crow.bin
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,21 +1,19 @@
 FROM ubuntu:xenial
-WORKDIR /home/dev
 
 RUN apt-get update && \
-    apt-get upgrade -y && \
-    apt-get install -y \
+    apt-get install -y --no-install-recommends software-properties-common && \
```
simonvanderveldt: @csboling I made some changes to the Dockerfile to make it a bit quicker since there is no layer caching on GitHub Actions (yet).

Difference in image size (top is after the changes)
```
$ d images | grep -i crow
crow-builder                           small               5ca08df689fc        About an hour ago   835MB
crow-builder                           latest              a5ae1f93f8a9        About an hour ago    923MB
```
It seems to be building the firmware correctly, but would probably be good to test if it still works as expected for you?

Dockerfile
```diff
@@ -31,4 +29,4 @@ RUN tar -xzf lua.tar.gz && \
 RUN luarocks install fennel
 
 WORKDIR /target
-CMD make
\ No newline at end of file
+ENTRYPOINT ["make"]
```
simonvanderveldt: I switched to `ENTRYPOINT` so we can easily pass arguments to `make`

.github/workflows/build-release.yml
```diff
@@ -0,0 +1,22 @@
+name: Build release firmware
+on:
+  release:
+    types: [created, edited]
```
simonvanderveldt: This should work for release although I couldn't find a way to test this

csboling:
Yep, works great! Thanks for wrestling with this.

PR #257, opened 2019-11-18
support i2c commands without arguments

trentgill:
This is untested because I don't think there are any devices that do this yet?


csboling:
I can confirm that this works for the [`MA.STEP` and `MA.RESET` commands for the Matrixarchate](https://github.com/monome/teletype/blob/ee402bfba6ac3015270e252bb50b65306131e952/src/ops/matrixarchate.c#L134).

PR #260, opened 2019-11-22
Support multiples of i2c-enabled devices

### LINKED ISSUE #255: i2c support multiple same-devices
At present, only TXi and er301 have support for multiple units. They both implement it in different ways: TXi uses ascending channel indexes. [1,4]=first device, [5-8]=second device. er301 adds 100 to the command to point to the second device.

I want to move toward a unified syntax for working with multiple devices (and one that feels natural in the lua env). This represents a slight departure from the teletype ops. I'm adding multiple JF unit support in the forthcoming update, so that device (and W/ shortly) will be added to this list, so i think a single syntax is important.

Syntax proposal:
```lua
-- current syntax. remains & always talks to the first / default address
ii.jf.trigger()

-- proposed new syntax. nb: only works if the device supports multiple devices
ii.jf[1].trigger() -- identical to above, but *required* when using multiple devices
ii.jf[2].trigger() -- speaks to the second defined address
```

To add support, the only change required in the i2c descriptor should be:
```lua
-- current. keep this for devices that only support one address
i2c_address = 0x70

-- proposed for muti-address capable devices
i2c_address = {0x70, 0x75} -- a list of addresses, in the order they will be indexed
```

trentgill:
Enables extended syntax for communicating with multiple devices that use the same i2c commands but send on different addresses. Forthcoming JF & W/ support will use the same.

examples:
```lua
ii.jf.trigger(1) -- trigger channel 1 on the default address (existing syntax is unchanged)
ii.jf[1].trigger(1) -- trigger channel 1 on the first device (same as first)
ii.jf[2].trigger(1) -- trigger channel 1 on the second device
ii.jf.trigger(1) -- trigger channel 1 on the last used device (second device)
```
The latching of the `ii.jf.` syntax is not ideal to me, but otherwise the implementation is substantially more complex (requiring metatables for every lookup). As a result, if a script uses more than 1 device, we should strongly suggest / describe-as-a-requirement, to use indexed module names eg `jf[1]` and `jf[2]`.

Getter syntax is the same, but the `ii.<module>.event` callback, optionally has an additional argument `index` which contains the index of the device that is returning data.

```lua
-- default callback for druid / norns / max
ii.jf.get('mode')
--> ^^ii.jf('mode',1,1) -- note the additional '1' at end which is device index
ii.jf[1].get('mode')
--> ^^ii.jf('mode',1,1)
ii.jf[2].get('mode')
--> ^^ii.jf('mode',1,2) -- here we return device 2
```

Multiple devices share a single event callback, but there's a new argument `index`:
```lua
ii.jf.event = function( event, value, index )
  if index == 1 then
    -- device 1
    if event == 'mode' then
      mymode = value
    end
  else
    -- device 2
  end
end
```

fixes #255 

TODO:
* test er301


trentgill:
Tested txi and it works fine with either the Teletype style sequential channel, or the crow style indexing:
```lua
ii.txi.get('param', 5)
-- same as
ii.txi[2].get('param', 1)
```

trentgill:
@tehn @dndrks I believe this PR is complete now. I've tested to the best of my ability without having a musical setup with multiple devices.

Added a very terse addition to the v1.0.3 docs PR. The cool thing is while this PR represents a lot of changes to the ii-descriptor code generation, the new user-facing syntax is just a tiny change! All that changed is the possibility of indexing an `ii.<device>`, and an additional argument to the event callback (which most people will never know existed!).

PR #261, opened 2019-11-25
delay() returns allocated metro and takes repeats

trentgill:
cc @csboling who had the idea to extend `delay` in this way.

`delay()` is updated to `delay( time, function, repeats )`.
'repeats' will cause the action to be performed more than 1 time. if it is omitted, defaults to 1 for standard delay behaviour.

Additionally, the delay() call returns the allocated metro object, so the user can call do:
```lua
dd = delay( 1.0, function(count) print(count) end, 5 )
-> 1
-> 2
-> 3
dd:stop() -- pauses execution of the delay
dd:start() -- resumes at current count (but waits the full delay time, regardless of when in the phase it was stopped)
-> 4
-> 5
print(dd)
-> table 0x200... -- this will refer to the metro that was created even after the delay is done.
```
This starts to become an alternate syntax for the metro system with the `repeats` really just reimplementing the `count` parameter to the metro lib. Key differences:
* no event happens immediately, instead waiting for one time-interval
* metro.init & metro:start are linked
* metro:free automatically cleans up when the sequence is complete

design question: should the arg list be (event, time, repeats) so it more closely echoes the metro system? ( think the answer is yes)


trentgill:
@tehn headsup - i just switched the arg order to match metro initialization.

tehn:
excellent!

tehn:
seems dumb but you could also give it a default `time` of 1.0s

trentgill:
Just checked and the metro lib already defaults to `t=1.0` so that will flow through to this if `t==nil`

PR #270, opened 2019-12-04
16n faderbank support

trentgill:
The ii descriptor is updated to allow command names that are numbers, such that you can use numerical iterators / math operations to decide which fader you are processing.

* Untested! If someone with a 16n could try this it would be greatly appreciated.
* I'm unsure who to attribute in the `manufacturer` string? `Lines community` or something?

Usage is:
```lua
-- request a fader value from [1,16]
ii.faders.get(1)

-- receive the requested slider value
ii.faders.event = function( slider, level, device )
    -- slider is [1,16]
    -- level is in Volts (?)
    -- device returns which i2c address you requested
end

-- request slider 4 from the second faderbank (you'll need to change the i2c address on your faderbank)
ii.faders[2].get(4)
-- results in same `ii.faders.event` above, but 'device' = 2
-- up to 4 faderbanks are supported
```


simonvanderveldt:
Does anyone of us have a 16n?

PR #276, opened 2019-12-06
CPU usage dramatically reduced in the ADC driver

### LINKED ISSUE #272: Remove delay() in the ADC driver
Currently, the DSP block takes up about 40% of CPU time, but the majority of that time is just burning cycles, waiting for the ADC to be ready to query after setting the NSS pin: https://github.com/monome/crow/blob/master/ll/ads131.c#L205

This is just initiating the ADC reading for the *next* dsp frame, so we don't need to wait. So rather than using a delay() in the DSP thread, just set a hardware timer (with very high priority), that calls the code that should happen after the delay. There's still 2 unused hardware timers (that have a different interface than the user-facing ones), so can just that.

This leaves almost double the CPU cycles for the main thread, meaning latency & jitter in the event queue should be drastically reduced. It also means there's ample CPU time for implementing more complex DSP processing (eg. #67 and #13 ). And/or could reduce samples-per-block (currently 32) which improves timing accuracy for audio rate ASLs, and increases sampling-rate for the ADC (currently ~1.5kHz).

trentgill:
Previously, the ADC driver had a blocking delay waiting for the ADS131 to have valid data to be clocked out in the SPI driver.

In this PR that delay is replaced with a timer that delays a continuation which requests samples for the next block. The wait time happens in the background.

The net result is ~35% less CPU usage. Specifically this increases the amount of time available for the DSP loop (so we can do more stuff!), and/or reduces latency & jitter for the event queue as the main loop has about 2x the cycles each time around. This is likely only noticeable with a CPU intensive lua script as the C side of the project never uses more than ~25% cpu.

Fixes #272

### DIFF
--- a/lib/io.c
+++ b/lib/io.c
@@ -11,10 +11,10 @@
 
 #define IN_CHANNELS ADDA_ADC_CHAN_COUNT
 
-void IO_Init( void )
+void IO_Init( int adc_timer_ix )
 {
     // hardware layer
-    ADDA_Init();
+    ADDA_Init(adc_timer_ix);
 
     // dsp objects
     Detect_init( IN_CHANNELS );
--- a/lib/io.h
+++ b/lib/io.h
@@ -2,7 +2,7 @@
 
 #include <stm32f7xx.h>
 
-void IO_Init( void );
+void IO_Init( int adc_timer_ix );
 void IO_Start( void );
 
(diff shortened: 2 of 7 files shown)


tehn:
@trentgill shall we merge some of these prior to 1.0.3?

PR #278, opened 2019-12-20
Syntax-check Lua files at compile time

trentgill:
Using `luac` with the `-p` option, this checks every lua file for syntax errors at compile time. This helps when making quick changes to the crow-lua-libs as typos are only caught by a text editor with linting, otherwise crow will probably just fail to boot into the lua env properly, erroring on the syntax error and abandoning the whole file.

Additionally, this removes a double-escape quote required by the lua -> c-header converter, by using the `[[`, `]]` string identifier syntax.

@simonvanderveldt I'm not sure if this requires any changes to the Actions setup, but I believe `luac` is part of the standard Lua install. Do changes to the Makefile need to be manually updated, or will the build process automatically use the updated Makefile?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,6 +270,7 @@ zip: $(BIN)
 	@$(FENNEL) --compile $< > $@
 
 %.lua.h: %.lua util/l2h.lua
+	@luac -p $<
 	@echo l2h $< "->" $@
 	@lua util/l2h.lua $<
 
--- a/lua/ii.lua
+++ b/lua/ii.lua
@@ -45,8 +45,7 @@ function ii_LeadRx_handler( addr, cmd, data )
     ii[name].event(ii[name].e[cmd], data, ix)
 end
 
--- NOTE: weird double-escaped quotes down here for the c compiler
-function ii.e( name, event, ... ) crow.tell('ii.'..name,'\\''..tostring(event)..'\\'',...) end
+function ii.e( name, event, ... ) crow.tell('ii.'..name,'[['..tostring(event)..']]',...) end
 
 ii.self =
     { cmds = { [1]='output'


simonvanderveldt:
> @simonvanderveldt I'm not sure if this requires any changes to the Actions setup, but I believe `luac` is part of the standard Lua install. Do changes to the Makefile need to be manually updated, or will the build process automatically use the updated Makefile?

The Action simply runs the Docker image built from the Dockerfile in this repo, which currently simply calls `make` https://github.com/monome/crow/blob/master/Dockerfile#L32

`luac` is installed, CI passed without issues.

It would be nice to add some logging so show what's happening since there's no feedback in the logs at the moment at all about what's happening.
I.e. I checked the logs and didn't see anything in there different compared to before.

trentgill:
Currently the `luac` command is silenced with the `@` prefix in make. What do you want to see printed? We could change the `l2h _.lua -> _.lua.h` to say something else? I always find build scripts way too noisy, so try and reduce everything to a single printed line per file processed (allows error messages to be seen more clearly IMO)

simonvanderveldt:
For me a simple `Checking syntax of Lua files` log line or something like that would already be a good improvement. Ideally with feedback that it passed when everything is OK.

I would probably let it output all the files it's checking so you can see/confirm that it's really doing/has scanned what you expected. Although I also understand the noise point you mentioned. Would that be too noisy for you?

What happens when something is wrong? Does it show the failures and does it exit with a non-zero exit code?

trentgill:
When the syntax check fails, it errors non-zero and the build fails. The `l2h` script on the other hand basically never fails so long as the file it's wrapping exists.

Is it important to print that a file was checked if it passes? If the file fails to parse, `luac` prints the syntax error in a very readable way to help the dev fix it.

Should I just move the `echo` line above the `luac` check so it's clear which file is building before a potential error message from luac?

simonvanderveldt:
> When the syntax check fails, it errors non-zero and the build fails.

:+1: 

> Is it important to print that a file was checked if it passes? If the file fails to parse, `luac` prints the syntax error in a very readable way to help the dev fix it.

In general I approach these validations as something that should be explicit in its output and not require the person who looks at the output to go in and look at the code to determine what is being validated/done. I.e. meaningful feedback.
You might know right now what it's doing, but in 6 months time? What if we have multiple directories containing Lua, does it only check one or both? Can I see that when it runs? Etc.

In the end I don't really care since I don't look at this a lot, but this would be my suggestions from a bunch of experience with CI/CD at work :)

> Should I just move the `echo` line above the `luac` check so it's clear which file is building before a potential error message from luac?

That could be one option, although it doesn't show that a syntax check/`luac -p` is being run on it of course.

Issue #264, opened 2019-11-27
ii.<module>.help() truncated to 1024chars


csboling:
With crow v1.0.2, in druid I see the following:

```
> ii.kria.help()

> print('****CAW!')
-- commands
ii.kria.preset( number )
ii.kria.pattern( number )
ii.kria.scale( number )
ii.kria.period( time )
ii.kria.position( track, param, pos )
ii.kria.loop_start( track, param, pos )
ii.kria.loop_length( track, param, pos )
ii.kria.reset( track, param )
ii.kria.mute( track, state )
ii.kria.toggle_mute( track )
ii.kria.clock( track )
-- request params
ii.kria.get( 'preset' )
ii.kria.get( 'pattern' )
ii.kria.get( 'scale' )
ii.kria.get( 'period' )
ii.kria.get( 'position', track, param )
ii.kria.get( 'loop_start', track, param )
ii.kria.get( 'loop_length', track, param )
ii.kria.get( 'reset', track )
ii.kria.get( 'mute', track )
ii.kria.get( 'cv', track )
-- then receive
ii.kria.event = function( e, data )
  if e == 'preset' then
    -- handle preset param here
  elseif e == 'pattern' then
  elseif e == 'scale' then
  elseif e == 'period' then
  elseif e == 'position' then
  elseif e == 'loop_start' then
  elseif e == 'loop_length' then
  elseif e == 'reset' then
  elseif e == 'mut****CAW!
```

This only currently affects ii.kria.help because the string is long enough that it can't be sent in a single `Caw_send_luachunk`. Not sure what the preferred way is to handle this in general.

trentgill:
Hrmmm... `Caw_send_luachunk` is limited only by the USB TX buffer (currently [1024bytes](https://github.com/monome/crow/blob/d6e0326c5b85c15dbe579a8d002a82b0b0d17728/usbd/usbd_cdc_interface.c#L55)).

The brute force solution is just allocating a bigger buffer (at the expense of runtime memory).

We could add some complex code to break a string into 1024byte chunks [here](https://github.com/monome/crow/blob/d6e0326c5b85c15dbe579a8d002a82b0b0d17728/lib/lualink.c#L338). This would require an arbitrary delay, waiting for the usb buffer to be cleared, or a callback from the usb tx driver, or some kind of queue system. This seems like massive overkill to me.

We could break up the `ii.<module>.help()` into separate parts. Or take an optional string argument like `'commands'` or `'events'` which would only print the part you were interested in.

[...]

PR #284, opened 2020-01-31
I2c support for txo

nordseele:
Adds ii support for one or multiple TXo modules (BPC Music Telex-o).
Most of the commands have been tested from both Druid and Max.

### DIFF
--- a/lua/ii/txo.lua
+++ b/lua/ii/txo.lua
@@ -457,7 +457,6 @@ do return
              , { 'state', s16 }
              }
     }
--- To do: eor, eoc, loop, bpol, cycl ??
   , { name = 'env_eoc'
     , cmd  = 0x6B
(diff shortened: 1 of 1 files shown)


tehn:
anyone able to test this?

csboling:
Tried most of the major commands, works great! As I suspected this is a lot of commands for how `ii.*.help()` works, so you don't see any help text after running `ii.txo.help()` until you force the USB write to flush, see #264

trentgill:
Merging this now.

@nordseele thanks so much for your efforts!!

Sidenote: This adds 8kB to the binary and now we only have about 60kB. It feels important to support the whole gamut of TXO commands than some arbitrary subset. If we get into a situation where we run out of space, we can use link time optimization to save ~50kB (at the expense of longer build times).

Issue #280, opened 2019-12-30
Failures in init() not reported


ngwese:
Using the released version 1.0.2 firmware I've noticed that runtime errors which occur within the `init()` function are not reported (in druid).

The lack of error reporting gives rise to hard to diagnose bugs due to the silent early termination of the `init()` function.

Reproduction steps:
- Create `init_fail.lua` with the following contents:
```
reset = true

function init()
  print("init start")
  reset(true)
  print("init end")
end

```
- Create `toplevel_fail.lua` with the following contents:
```
reset = true

function my_init()
  print("init start")
  reset(true)
  print("init end")
end

my_init()
```
- In `druid` run `r init_fail.lua` and compare that to `r toplevel_fail.lua`

The expected behavior is that failures which occur in `init()` when invoked by the firmware are reported.

trentgill:
Will investigate this for 1.0.4 - thanks for the easy repro script :)

PR #286, opened 2020-03-08
store script name and running state to show correct name

### LINKED ISSUE #283: After a `r` from druid, crow reports the wrong script as 'running'
See: https://llllllll.co/t/crow-help-druid/25864/184

Basically issue is after an `r` command from druid, crow restarts the lua VM and prints the 'current script'. At present the `r` command is run directly from RAM, versus a regular boot which reads from flash.

Need to add an extra case for when the module 'restarts' but is running a script in RAM, and thus read the script name from that RAM buffer.

csboling:
Shows the name of the script that's running when running a script from RAM.

Fix #283 

crow-scriptname

In the `^^e` case the script buffer is immediately freed after evaluation so we need to store the script name, as well as a flag indicating that the running script name takes precedence over printing the entire script contents if a user script is already uploaded. Rather than adding more complexity to `REPL_script_name` depending on this extra state, script evaluation was refactored into a `REPL_run_script` function which always copies the appropriate script name to `running_script_name` once evaluation succeeds, and then `REPL_script_name` can always just send out this buffer. The `REPL_run_script` function could also be a good place for e.g. testing `init` as in https://github.com/monome/crow/issues/280.

### DIFF
--- a/lib/repl.c
+++ b/lib/repl.c
@@ -36,14 +36,14 @@ void REPL_init( lua_State* lua )
 
     switch( Flash_which_user_script() ){
         case USERSCRIPT_Default:
-	    REPL_run_script( USERSCRIPT_Default, NULL, 0 );
+            REPL_run_script( USERSCRIPT_Default, NULL, 0 );
             break;
         case USERSCRIPT_User:
         {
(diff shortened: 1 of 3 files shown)


tehn:
rad!! thank you!

PR #287, opened 2020-03-21
add error handler for lua_pcall

### LINKED ISSUE #228: Better error messages over USB
Presently crow sends the default error message which looks like:
```lua
> print(jfd'
[string "eval"]:1: unfinished string near '''
```
The issue is we're using `dostring` in Lua (because there's no filesystem), and as a result we don't get linebreaks, because it's just a long string.

For errors from druid, it would be beneficial to simply strip the leading `[string "eval"]:1:` and replace with something like `Error: ` to make it clear that crow can't evaluate your command.

When it comes to running scripts, it's more important to be able to find the real 'line' in the source that causes the issue. Is it possible to manually iterate through the source file, splitting on newlines, to find the offending code. Kind of depends on how the Lua VM internally deals with a big string with multiple executable sections inside.

csboling:
Fix #228 

Adds an error handler for the `lua_pcall` invocation, patterned after what [`docall`](https://github.com/trentgill/lua/blob/9d2732aaab53926662f352e347484a92f7afa0bd/src/lua.c#L176) does.

Example:
```
 <crow connected>

> print(5)
5

> print(jfd'
[string "eval"]:1: unfinished string near '''

> r lua/bad.lua
running lua/bad.lua
[string "eval"]:10: attempt to call a nil value (global 'boom')
stack traceback:
  [string "eval"]:10: in function 'three'
  [string "eval"]:6: in function 'two'
  [string "eval"]:2: in function 'one'
  [string "eval"]:13: in main chunk
User script evaluation failed.
```
where `bad.lua` is:
```lua
 1  function one()
 2   two()
 3  end
 4
 5  function two()
 6    three()
 7  end
 8
 9  function three()
10    boom()
11  end
12
13  one()
```

Notes:
* need to `goto` the "send error" code if `luaL_loadbuffer` failed due to a syntax error, otherwise we would always get "attempt to call a string value" instead of the actual error message in repl mode
* still always shows the chunk name as `[string "eval"]:`. From [here](https://github.com/trentgill/lua/blob/9d2732aaab53926662f352e347484a92f7afa0bd/src/lua.c#L335) I think possibly passing "=eval" rather than "eval" to `luaL_loadbuffer` would give `eval:1:...`

### DIFF
--- a/lib/lualink.c
+++ b/lib/lualink.c
@@ -493,28 +493,54 @@ static void Lua_linkctolua( lua_State *L )
     }
 }
 
+static int Lua_handle_error( lua_State *L )
+{
+    const char *msg = lua_tostring( L, 1 );
+    if( msg == NULL ){
+        if( luaL_callmeta( L, 1, "__tostring" )
+	 && lua_type ( L, -1 ) == LUA_TSTRING ) {
+	    return 1;
+	}
+	else {
+	    msg = lua_pushfstring( L
+				 , "(error object is a %s value)"
+				 , luaL_typename( L, 1 ) );
+	}
+    }
+    luaL_traceback( L, L, msg, 1 );
+    return 1;
+}
+
 uint8_t Lua_eval( lua_State*     L
                 , const char*    script
                 , size_t         script_len
                 , ErrorHandler_t errfn
                 ){
-    int error;
-    if( (error = luaL_loadbuffer( L, script, script_len, "eval" )
-              || lua_pcall( L, 0, 0, 0 )
-        ) ){
-        //(*errfn)( (char*)lua_tostring( L, -1 ) );
-        Caw_send_luachunk( (char*)lua_tostring( L, -1 ) );
-        lua_pop( L, 1 );
-        switch( error ){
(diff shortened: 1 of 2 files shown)


trentgill:
This looks great! Will try and test it today!

csboling:
Looking at your watchdog changes I realized that this just wraps the initial eval, not all the places event handlers might call into user code. Should be possible to factor this out to a `docall` of our own that can be used anywhere user code is invoked.

csboling:
I'm gonna close this and rework it based on changes and discussion in #288 , #290

trentgill:
Yeah this is weird because the event handlers are calling functions in the Lua env, while the initial bootstrap & loading the userscript is adding a bunch of new Lua code into the environment before executing.

I'd love to hear how we could use `docall` but i'm not totally clear how it solves this problem.

In this PR you've already broken up the `loadfile` from the `pcall`, so we should likely just split into 2 functions for these parts, then everything can share the `pcall_protected` function?

Really excited to have this improved error handler! It's something I've been wanting to do for ages but didn't know where to start!

PR #288, opened 2020-03-21
Watchdog timeout for long running scripts

trentgill:
Using lua's `sethook` functionality to run a test every 2 seconds (this time can change), to make sure it's not stuck in an infinite loop. Obviously it's theoretically possible that a user script is taking this long, but we can assume that's likely not realistic & assume there is a bug in the script causing an infinite loop.

This should solve the situation where someone uploads a script that infinite-loops in the `init` function and requires a force-bootloader to become responsive again.

Unfortunately, this doesn't work for infinite recursions, because the lua stack overflows and crashes crow well before 2s is up.

I'm still not sure it solves all cases where a userscript can cause an unresponsive-on-boot situation (likely issues related to i2c?) but i don't have reliable test scripts that cause this crash. If you do, please start a new issue!

Thanks to @samdoshi for the idea & pointer to using lua's debug hooks!

Also, @csboling is likely interested as it touches close to the error handling things.

### DIFF
--- a/lib/lualink.c
+++ b/lib/lualink.c
@@ -38,8 +38,8 @@
 
 #include "build/ii_lualink.h" // generated C header for linking to lua
 
-#define WATCHDOG_COUNT 2
-#define WATCHDOG_FREQ  0x7FFFF // ~500ms
+#define WATCHDOG_FREQ      0x100000 // ~1s how often we run the watchdog
+#define WATCHDOG_COUNT     2        // how many watchdogs before 'frozen'
 
 const struct lua_lib_locator Lua_libs[] =
     { { "lua_crowlib"   , lua_crowlib   }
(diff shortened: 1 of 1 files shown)


csboling:
> doesn't work for infinite recursions

As far as I can tell the only check for stack overflow is based on [LUAI_MAXSTACK](https://github.com/trentgill/lua/blob/9d2732aaab53926662f352e347484a92f7afa0bd/src/luaconf.h#L715), so memory is exhausted before a Lua error gets raised. It seems there [used to be](https://www.lua.org/source/5.0/llimits.h.html) a `LUA_MAXCALLS` for this (255 by default). We could either change the `LUAI_MAXSTACK` to the amount of available interpreter memory, or maybe use `lua_sethook` to register a call hook as well that uses [`lua_getstack`](https://www.lua.org/manual/5.3/manual.html#lua_getstack) to manually walk up the call stack and see if we've exceeded some call depth.

lib/lualink.c
```diff
@@ -72,6 +77,7 @@ lua_State* Lua_Init(void)
     L = luaL_newstate();
     luaL_openlibs(L);
     Lua_linkctolua(L);
+    lua_sethook(L, hook, LUA_MASKCOUNT, WATCHDOG_FREQ);
```
csboling: I think this `lua_sethook` is redundant because `Lua_eval` uses `lua_pcall_protected`

trentgill: spot on! initially this was just to setup the hook, but then i ended up having to reset it each time a new external call happened. pushed the fix.

PR #291, opened 2020-03-23
error handler for reporting tracebacks from all user code

### LINKED ISSUE #228: Better error messages over USB

csboling:
Like https://github.com/monome/crow/pull/287, but wraps all calls into user code. Includes hooks for both watchdog timeout in https://github.com/monome/crow/pull/288 and a protection against exceeding 255 call frames.

Fix #228 

Example:
```
> print(jfd'
repl:1: unfinished string near '''

> function f() f() end; f()
Call depth exceeded 255
repl:1: recursion limit exceeded
stack traceback:
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  ...
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in main chunk

> input[1].change = f

> input[1].mode = 'change'
Call depth exceeded 255
repl:1: recursion limit exceeded
stack traceback:
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  ...
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  repl:1: in function 'f'
  [string "local Input = {}..."]:93: in function 'change_handler'
error: input change
```

Notes

* since the shared function changed to a different signature than `pcall` and is used to wrap invocations of user code, named it `Lua_call_usercode` for consistency with other crow-style `Lua_` functions rather than lua's `lua` and `luaL` prefixes
* instead of `[string 'eval']`, shows whether the code is from the `repl` or the `userscript`. Changed `Lua_eval` to accept this chunk name as a parameter instead of the unused `ErrorHandler_t`.
* I'm not sure if this recursion detection is too aggressive. It's based on walking up the call stack as `luaL_traceback` does to determine if we're more than 255 calls deep. In particular due to tail call optimization, function calls might not grow the stack, needs more testing to find if there are situations where this would be an arbitrary restriction.
* both the `Caw_send_luachunk` to report the error to the host and the `_printf` have moved inside the error handler, ensuring that all error reporting happens from one place. Since the traceback now shows you where the error happened (`function 'change_handler'`) it may not be necessary to also `Caw_send_luachunk("error: input change")` etc.

### DIFF
--- a/lib/lualink.c
+++ b/lib/lualink.c
@@ -69,6 +69,16 @@ static int Lua_handle_error( lua_State* L );
 static void timeouthook( lua_State* L, lua_Debug* ar );
 static void callhook( lua_State* L, lua_Debug* ar );
 
+// Handler prototypes
+void L_handle_toward( event_t* e );
+void L_handle_metro( event_t* e );
+void L_handle_in_stream( event_t* e );
+void L_handle_change( event_t* e );
+void L_handle_ii_leadRx( event_t* e );;
+void L_handle_ii_followRx( event_t* e );
+void L_handle_ii_followRx_cont( uint8_t cmd, int args, float* data );
+void L_handle_midi( event_t* e );
+
 void _printf(char* error_message)
 {
     printf("%s\n",error_message);
@@ -612,15 +622,15 @@ static int Lua_call_usercode( lua_State* L, int nargs, int nresults )
 // Public Callbacks from C to Lua
 void L_queue_toward( int id )
 {
-    event_t e = { .type    = E_toward
+    event_t e = { .handler = L_handle_toward
                 , .index.i = id
                 };
     event_post(&e);
 }
-void L_handle_toward( int id )
+void L_handle_toward( event_t* e )
 {
     lua_getglobal(L, "toward_handler");
-    lua_pushinteger(L, id+1); // 1-ix'd
+    lua_pushinteger(L, e->index.i + 1); // 1-ix'd
     if( Lua_call_usercode(L, 1, 0) != LUA_OK ){
         Caw_send_luachunk("error running toward_handler");
         Caw_send_luachunk( (char*)lua_tostring(L, -1) );
@@ -631,17 +641,17 @@ void L_handle_toward( int id )
 
 void L_queue_metro( int id, int state )
 {
-    event_t e = { .type    = E_metro
+    event_t e = { .handler = L_handle_metro
                 , .index.i = id
                 , .data.i  = state
                 };
     event_post(&e);
 }
-void L_handle_metro( const int id, const int stage)
+void L_handle_metro( event_t* e )
 {
     lua_getglobal(L, "metro_handler");
-    lua_pushinteger(L, id+1 -2); // 1-ix'd, less 2 for adc rebase
-    lua_pushinteger(L, stage+1); // 1-ix'd
+    lua_pushinteger(L, e->index.i +1 -2); // 1-ix'd, less 2 for adc rebase
+    lua_pushinteger(L, e->data.i +1);     // 1-ix'd
     if( Lua_call_usercode(L, 2, 0) != LUA_OK ){
         Caw_send_luachunk("error running metro_handler");
         Caw_send_luachunk( (char*)lua_tostring(L, -1) );
@@ -651,17 +661,17 @@ void L_handle_metro( const int id, const int stage)
 
 void L_queue_in_stream( int id )
 {
-    event_t e = { .type    = E_stream
+    event_t e = { .handler = L_handle_in_stream
                 , .index.i = id
                 , .data.f  = IO_GetADC(id)
                 };
     event_post(&e);
 }
-void L_handle_in_stream( int id, float value )
+void L_handle_in_stream( event_t* e )
 {
     lua_getglobal(L, "stream_handler");
-    lua_pushinteger(L, id+1); // 1-ix'd
-    lua_pushnumber(L, value);
+    lua_pushinteger(L, e->index.i +1); // 1-ix'd
+    lua_pushnumber(L, e->data.f);
     if( Lua_call_usercode(L, 2, 0) != LUA_OK ){
         Caw_send_luachunk("error: input stream");
         Caw_send_luachunk( (char*)lua_tostring(L, -1) );
@@ -671,17 +681,17 @@ void L_handle_in_stream( int id, float value )
 
 void L_queue_change( int id, float state )
 {
-    event_t e = { .type    = E_change
+    event_t e = { .handler = L_handle_change
                 , .index.i = id
                 , .data.f  = state
                 };
     event_post(&e);
 }
-void L_handle_change( int id, float state )
+void L_handle_change( event_t* e )
 {
     lua_getglobal(L, "change_handler");
-    lua_pushinteger(L, id+1); // 1-ix'd
-    lua_pushnumber(L, state);
+    lua_pushinteger(L, e->index.i +1); // 1-ix'd
+    lua_pushnumber(L, e->data.f);
     if( Lua_call_usercode(L, 2, 0) != LUA_OK ){
         printf("ch er\n");
         Caw_send_luachunk("error: input change");
@@ -692,19 +702,19 @@ void L_handle_change( int id, float state )
 
 void L_queue_ii_leadRx( uint8_t address, uint8_t cmd, float data )
 {
-    event_t e = { .type   = E_ii_leadRx
-                , .data.f = data
+    event_t e = { .handler = L_handle_ii_leadRx
+                , .data.f  = data
                 };
     e.index.u8s[0] = address;
     e.index.u8s[1] = cmd;
     event_post(&e);
 }
-void L_handle_ii_leadRx( uint8_t address, uint8_t cmd, float data )
+void L_handle_ii_leadRx( event_t* e )
 {
     lua_getglobal(L, "ii_LeadRx_handler");
-    lua_pushinteger(L, address);
-    lua_pushinteger(L, cmd);
-    lua_pushnumber(L, data);
+    lua_pushinteger(L, e->index.u8s[0]); // address
+    lua_pushinteger(L, e->index.u8s[1]); // command
+    lua_pushnumber(L, e->data.f);
     if( Lua_call_usercode(L, 3, 0) != LUA_OK ){
         printf("!ii.leadRx\n");
         Caw_send_luachunk("error: ii lead event");
@@ -715,11 +725,12 @@ void L_handle_ii_leadRx( uint8_t address, uint8_t cmd, float data )
 
 void L_queue_ii_followRx( void )
 {
-    event_t e = { .type = E_ii_followRx };
+    event_t e = { .handler = L_handle_ii_followRx };
     event_post(&e);
 }
-void L_handle_ii_followRx( void )
+void L_handle_ii_followRx( event_t* e )
 {
+    // FIXME: should pass the 'cont' as a fnptr continuation
     ii_process_dequeue_decode();
 }
 void L_handle_ii_followRx_cont( uint8_t cmd, int args, float* data )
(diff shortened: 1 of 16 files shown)


lib/lualink.c
```diff
@@ -537,28 +543,71 @@ void Lua_crowbegin( void )
 {
     printf("init()\n"); // call in C to avoid user seeing in lua
     lua_getglobal(L,"init");
-    if( lua_pcall_protected(L,0,0,0) != LUA_OK ){
-        Caw_send_luachunk( (char*)lua_tostring( L, -1 ) );
-        lua_pop( L, 1 );
+    if( Lua_call_usercode(L,0,0) != LUA_OK ){
+        Caw_send_luachunk("error running init()");
```
trentgill: This line is a reversion. It changed to the `lua_tostring` call in a recent commit to get the actual error message from the lua env.

lib/lualink.c
```diff
@@ -537,28 +543,71 @@ void Lua_crowbegin( void )
...
+
+// Check for too much recursion
+static void callhook( lua_State* L, lua_Debug* ar )
+{
+    for (int level = 0; lua_getstack(L, level, ar); level++) {
+        if (level > MAX_CALLFRAMES) {
+	    Caw_send_luachunk("Call depth exceeded " TOSTRING(MAX_CALLFRAMES));
+	    luaL_error(L, "recursion limit exceeded");
+	}
     }
```
trentgill: Maybe I'm misunderstanding this logic, but couldn't this just be:
```
if( lua_getstack(L, level, ar) > MAX_CALLFRAMES ){
  // recursion limit exceeded
}
```

lib/lualink.c
```diff
@@ -537,28 +543,71 @@ void Lua_crowbegin( void )
...
+}
+
+static int Lua_call_usercode( lua_State* L, int nargs, int nresults )
 {
-    lua_sethook(L, hook, LUA_MASKCOUNT, WATCHDOG_FREQ); // reset hook
-    watchdog = WATCHDOG_COUNT; // reset hook counter
-    return lua_pcall(L, nargs, nresults, errFunc);
+    lua_sethook(L, timeouthook, LUA_MASKCOUNT, WATCHDOG_FREQ); // reset timeout hook
+    watchdog = WATCHDOG_COUNT; // reset timeout hook counter
+    lua_sethook(L, callhook, LUA_MASKCALL, 0); // reset call hook
```
trentgill: Is there a potential performance penalty here if we're calling the hook everytime lua makes a function call? I'm not even sure how to benchmark this, but it seems especially time-intensive if we have to iterate over the whole lua call stack every time lua enters a new function.

lib/lualink.c
```diff
@@ -537,28 +543,71 @@ void Lua_crowbegin( void )
...
+
+    int errFunc = lua_gettop(L) - nargs;
+    lua_pushcfunction( L, Lua_handle_error );
+    lua_insert( L, errFunc );
+    int status = lua_pcall(L, nargs, nresults, errFunc);
+    lua_remove( L, errFunc );
+
+    lua_sethook(L, timeouthook, 0, 0);
+    lua_sethook(L, callhook, 0, 0);
```
trentgill: Just curious why these need to be unset after ending the code execution? Is it just a sanitation thing?

lib/lualink.c
```diff
@@ -40,6 +40,9 @@
 
 #define WATCHDOG_FREQ      0x100000 // ~1s how often we run the watchdog
 #define WATCHDOG_COUNT     2        // how many watchdogs before 'frozen'
+#define MAX_CALLFRAMES     255      // call depth before 'stack overflow'
+#define _TOSTRING(s) #s
```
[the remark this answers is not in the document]
csboling: `#s` in a macro is the "stringification operator", which replaces the argument with a string constant containing the text of the argument. Since we want to call `TOSTRING` on a macro value `MAX_CALLFRAMES` this requires a double expansion. https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html

This is just to give a string further down that says `"Call stack exceeded max depth of " TOSTRING(MAX_CALLFRAMES)`. Could easily hardcode this or just not report the depth.

lib/lualink.c
csboling: The Lua error always gets reported with `Caw_send_luachunk` and `_printf` in `Lua_handle_error`, so sending it here would wind up printing it twice.

lib/lualink.c
csboling: Yeah this function is probably actually quadratic in the number of call frames! Should just do `if( lua_getstack(L, MAX_CALLFRAMES, ar) ) != 0` -- `lua_getstack` already walks a linked list of function calls and returns 0 if there is no such level.

lib/lualink.c
csboling: Yeah just habitually cleaning up the state, don't think it matters, though it would for the `callhook` if we only want to install it for calling `init()`

csboling:
> amazed nobody has run into this on the norns... perhaps worth doing something similar in matron

Looks like matron [does](https://github.com/monome/norns/blob/master/matron/src/lua_eval.c#L108) exactly what the off-the-shelf Lua interpreter [does](https://github.com/lua/lua/blob/master/lua.c#L109). Basically the only difference here is a manual recursion check, "desktop" Lua environments seem to just let you recurse until you run out of stack space. On crow the stack (currently) grows until A Bad happens.

csboling:
> the user can't interact with the device to clear the script, which is always the result of a bad `init` function in the userscript

I think you could also crash the module by having code that doesn't terminate or grows the stack too large in a metro, input stream handler, or just freestanding code in the script outside any function.

lib/lualink.c
trentgill: Ahh cool. I'm still thinking in the previous mental model where the 'call' doesn't handle errors directly.

trentgill:
> > amazed nobody has run into this on the norns... perhaps worth doing something similar in matron
> 
> Looks like matron [does](https://github.com/monome/norns/blob/master/matron/src/lua_eval.c#L108) exactly what the off-the-shelf Lua interpreter [does](https://github.com/lua/lua/blob/master/lua.c#L109). Basically the only difference here is a manual recursion check, "desktop" Lua environments seem to just let you recurse until you run out of stack space (configured by [`LUAI_MAXSTACK`](https://github.com/lua/lua/blob/e4607523234f16ed9ed0436340b9315377dbfe7f/luaconf.h#L724)). On crow the stack (currently) grows until A Bad happens, since Lua doesn't know the size of available device memory.

I just had a quick play with `LUAI_MAXSTACK` and found that it's actually doing exactly what `callhook` is doing in this PR. I played around for a moment and it seems 600~650 is the max stack before 'A Bad' happens (at least in an empty VM).

Should we fall back onto the native lua support for this then? Some thoughts:
* the error is different, reading "lib:9: stack overflow" instead of "call stack depth exceeded"
* how do we change LUAI_MAXSTACK? do we need to fork lua/ and change luaconf.h?
* should the limit be lower? 512? 256? it seems kinda arbitrary at that depth, so less is better?

I still don't understand how this relates to the maximum space lua can take up in memory. I'm not sure why the system hardfaults without this limit. Stack overflow I get, but what about lua's dynamic allocations? Is there any way to give a lua a limit of how many kB it's allowed to use overall?

trentgill:
I've been playing with this a little bit now & it's so much nicer than before! Can't wait for folks to have this way-more-informative error messaging :) Feels like a solid 1.0.4 is coming together quickly!

csboling:
>     * the error is different, reading "lib:9: stack overflow" instead of "recursion limit exceeded"

"recursion limit" is just from a `lua_error` call in `Lua_handle_error` because you didn't actually run out of stack space, we cut you off.

>     * how do we change LUAI_MAXSTACK? do we need to fork lua/ and change luaconf.h?

Bummer, guess so. If it had an `#if defined()` we should just be able to pass `-DLUAI_MAXSTACK xxx` as a compiler flag. 

> I still don't understand how this relates to the maximum space lua can take up in memory. I'm not sure why the system hardfaults without this limit.

Yeah after reading some more code and documentation I'm not sure either. Everything runs in protected mode so allocation failures should just return `LUA_ERRMEM` from `lua_pcall` (unless the `setjmp` protection mechanism isn't working?). I don't have the UART set up but `Caw_send_luachunk` doesn't write anything, it seems it faults inside the `pcall`. Controlling allocation failure and limiting the stack size are the only means for controlling heap size that I've found.

trentgill:
Yeah, in general, I think `malloc` isn't catching the out of memory error (i've seen it on other projects of mine), but I dont have the requisite knowledge to know where to look / how to approach solving that issue. I guess that's a good learning / research project in these current times... I'll make a separate issue for that now.

Do you want to pull out the recursion testing from this PR and I'll make a new one with the forked `lua` repo? Would love to get this merged (I did a quick conflict resolution bc the event handler configuration changed).

Final question - in all the event handlers at the end of `lualink.c` they check for errors from the `Lua_call_usercode` and then call `send_lua_chunk` and `lua_pop`. Should those be removed as the error's already been reported in the `call_usercode` and this would double-pop the stack?

csboling:
>  pull out the recursion testing from this PR

Hook removed in [`cce598c`](https://github.com/monome/crow/pull/291/commits/cce598c0110faf9ee8f255752fadca67a86cea90).



> Final question - in all the event handlers at the end of `lualink.c` they check for errors from the `Lua_call_usercode` and then call `send_lua_chunk` and `lua_pop`. Should those be removed as the error's already been reported in the `call_usercode` and this would double-pop the stack?

I'm trying to mimic the behavior of `docall` here a bit which seems to make it the caller's responsibility to pop the message off the stack in case of an error, in case the caller wants to do something else with it. This is a little at odds with how the errors get sent out inside the error handler - the traceback gets sent out the USB port and the UART inside the errorhandler function, whereas Lua's `docall` deals with this after the `pcall` in `report`:

```c
static int report (lua_State *L, int status) {
  if (status != LUA_OK) {
    const char *msg = lua_tostring(L, -1);
    l_message(progname, msg);
    lua_pop(L, 1);  /* remove message */
  }
  return status;
}

static int dochunk (lua_State *L, int status) {
  if (status == LUA_OK) status = docall(L, 0, 0);
  return report(L, status);
}
```

I do not quite get how this handles the stack because it seems to me the error handler can put at least one additional value on the stack (the traceback, and an additional message if the error object has no `__tostring` method. Having the caller `lua_pop` here seems consistent with the usage in `lua.c`'s REPL implementation. I think maybe the additional "error: init()" or "error: input stream" feels redundant since you can see that the call stack started from e.g. `input_handler`, which feels a bit more like the "stock" Lua interpreter to me. Removed these messages in [`
ded616d`](https://github.com/monome/crow/pull/291/commits/ded616d1ab50c1ac58586dc7483dc3937dd63782).

trentgill:
Amazing! Thank you! Merged & merging the lua library change.

Issue #14, opened 2018-09-24
ADC interface methods


tehn:
- [x] finish design
- [ ] implement

_design notes_

- each input is independent, can be set up with different attributes
- basic access via QUERY ie `get_adc(1)` returns `adc(1,value)`
- POLL. set an interval. sends data at this interval. start/stop. (hard timer for each of these)
- HYSTERESIS/window/comparator. send an event based on a range/threshold setting, positive/negative/both edges. aka onCHANGE
- _quantization? could/should the hysteresis be abstracted outside of just the onCHANGE reporting?_
- _filter? slew/lag the input ie, become an envelope follower?_

trentgill:
```lua
input[n].mode( 'window', {list}, 0.03 ) -- calls input[n].window( ix or value? )
-- what should the arg be? index or value? return the number that we just went *above*?
-- what value to print when below the first item?

input[n].mode( 'scale', {note-list}, divs-per-scale, scale-size )
-- need to make the windows centered on the note, rather than above/below, default to chromatic? octaves?
-- divs-per-scale sets range of numbers accepted in {note-list}, default to 12
-- scale-size sets volts to wrap over, default to 1.0

-- the scale event's param returns the note division + divs-per-scale * range
-- chromatic @0V = 0, chromatic @1V=12, chromatic @2V=24

-- chromatic quantizer:
input[n].mode('scale')
input[n].scale = function(note) print(note) end

-- octave quantizer:
input[n].mode('scale',{0})

-- major scale:
input[n].mode('scale',{0,2,4,5,7,9,11})

-- buchla 12TET
input[n].mode('scale',{0,1,2,3,4,5,6,7,8,9,10,11},12,1.2)

-- extended major arpeggio
input[n].mode('scale', {0,4,7,11,14,17,21},24,2.0)

-- 19TET chromatic
input[n].mode('scale',{},19)
-- empty list uses default (taking divs into account)
```

trentgill:
```lua
-- input modes
'none'
'stream'
'change'

'window' -- list of voltage windows for windowed comparator
'scale' -- encapsulates 12TET-scale/quantize/ji with same syntax as output.scale
'peak' -- detect amplitude event
'volume' -- stream amplitude values
'freq' -- detects zero-crossings with abs-voltage-hysteresis. waits for a preset settling time
```

trentgill:
[...]

PR #298, opened 2020-03-28
'scale' option for Inputs (ie quantizing)

trentgill:
Furthers #14 by adding the 'scale' option.

Same structure as the output scaler/quantizer/shaper in #292 

```lua
-- set input 1 to quantize based on a major pentatonic scale, 12TET, 1v/8ve
input[1].mode( 'scale', {0,2,4,7,9} )
input[1].scale = function( note )
  -- this is called any time the input moves to a new note
end

-- shortcut for chromatic scaling
input[1].mode( 'scale', {} )

-- n-TET support, eg: 19TET pentatonic
input[1].mode( 'scale', {0,3,6,9,11}, 19 )

-- non-v/8 support, eg: buchla pitch
input[1].mode( 'scale', {0,2,4,7,9}, 12, 1.2 )

-- non-v/8 support, eg: major arpeggio over 2 octaves
input[1].mode( 'scale', {0,4,7,11,14,17,21}, 12, 2.0 )
```

Incomplete:
* Just intonation support
* Better testing
* Event callback type signature

Here are some ideas for the event callback type signature with some initial thoughts on pros/cons of the different choices. As always, it should come down to how these will integrate into a script. I'm happy to add a couple helper functions to the lua input library to allow conversion to different use-cases. eg: we can convert from `note` to `volts` very easily with an Input method.

```lua
--option1
input[1].scale = function( note )
  -- octave & note combined to a linear range where 0 == 0V
  -- notes would be integers like MIDI notes
  -- should they be sequential per the scale, or absolute like MIDI notes (ie. have gaps)
  -- eg sequential with scale {0,2,4,7,9} -> [1,2,3,4,5]
  -- eg absolute with same scale {0,2,4,7,9} -> [0,2,4,7,9
end

--option2
input[1].scale = function( note, octave )
  -- separate note & octave
  -- can easily achieve option1 with: octave+note/12 (or whatever nTET setting)
  -- is it useful to have them separate? i think not?
end

--option3
input[1].scale = function( volts )
  -- just pass the quantized voltage that has been landed at
  -- this feels more like a voltage processor where we're just cleaning up values
  -- eg: scale {0,2,4,7,9} -> [0.0, 0.16, 0.33, 0.55, 0.7]
  -- unfortunately you can't use the value as a table-lookup
  -- seems easier to make these values by using options1/2 and the input scale table
end
```


trentgill:
This is complete now & ready to merge. Just needs documentation update. Setup a channel as follows:

```lua
-- set input 1 to quantize based on a major pentatonic scale, 12TET, 1v/8ve
input[1].mode( 'scale', {0,2,4,7,9} )
input[1].scale = function( note )
  -- this is called any time the input moves to a new note
end

-- shortcut for chromatic scaling
input[1].mode( 'scale', {} )

-- n-TET support, eg: 19TET pentatonic
input[1].mode( 'scale', {0,3,6,9,11}, 19 )

-- non-v/8 support, eg: buchla pitch
input[1].mode( 'scale', {0,2,4,7,9}, 12, 1.2 )

-- non-v/8 support, eg: major arpeggio over 2 octaves
input[1].mode( 'scale', {0,4,7,11,14,17,21}, 12, 2.0 )
```

And the following event will be called:
```lua
input[1].scale = function( s )
  --[[ where:
    s = { index = _ -- the index of the note currently playing, wraps per 3rd arg to mode
        , octave = _ -- counts index rollover. typically represents which octave
        , note = _ -- looks up index in the provided note table & adds octave. akin to midinote in 12TET
        , volts = _ -- the exact voltage of the selected note for direct output
  ]]
end
```

And a few examples of why these are interesting
```lua
-- quantizer with portamento defined by .slew
output[1].slew = 0.1
input[1].scale = function(s)
  output[1].volts = s.volts
end

-- use index & octave to invert motion in a scale
myscale = {0,2,4,7,9}
input[1].mode( 'scale', myscale )
input[1].scale = function(s)
  output[1].volts = myscale[#myscale - s.index + 1]/12 + s.octave
end

-- use note for sending values to host / norns etc
input[1].mode( 'scale' )
input[1].scale = function(s)
  crow.tell( 'scale', 1, s.note ) -- this is the default behaviour if no function is declared
end
```

These examples suggest that `s.volts` is best if you just want to quantize an input stream and treat the values as notes directly.

`s.note` is best if you want to share the note selection with other systems, or if the scripter is already familiar with manipulating midi note numbers (or has scripts to do it).

`s.index` and `s.octave` are typically used together and are best if you want to do more complex transpositions / translations of notes. These values can also be used for more general control voltage manipulations. `s.octave` is similar to `'window'` with an additional inner set of panes in `s.index`. In that regard it can be useful to use the 'chromatic' behaviour with an empty table as 2nd argument:

`input[1].mode( 'scale', {}, panes_per_window, volts_per_window )`

PR #299, opened 2020-03-31
implement _sbrk syscall to prevent hard fault in malloc

### LINKED ISSUE #296: malloc() causes crash when out of memory
It should just return `NULL` and we handle the error in the code. This has the roll-on effect that Lua will crash the system, rather than failing to continue.

Perhaps something to do with the configuration in the root directory files? Maybe a linker issue where you declare minimum stack/heap sizes.

This has always perplexed me, so any guidance on how `malloc` determines whether there is free space in an embedded environment would be super helpful!

csboling:
Fix #296 

libc expects you to provide an implementation of the [`sbrk`](https://www.gnu.org/software/libc/manual/html_node/Resizing-the-Data-Segment.html#Resizing-the-Data-Segment) syscall in order for malloc to be able to determine when there is no more heap space.

Implementation is lifted from the ST examples, e.g. [here](https://github.com/trentgill/STM32_Cube_F7/blob/2147c24a0cbccfae331ca18a519f8ff6b91e860f/Projects/STM32F722ZE-Nucleo/Applications/USB_Host/HID_Standalone/SW4STM32/syscalls.c#L104), adapted to use GNU assembler syntax.

Test snippet:

```c
    Caw_send_luachunk("start\n");
    HAL_Delay(3);
    void* p;
    while (p = malloc(0x1000))
    {
        Caw_send_luachunk("alloc\n");
	HAL_Delay(3);
    }
    Caw_send_luachunk("fail\n");
    HAL_Delay(3);
```

Without this patch crow will hard fault and never print "fail", after this patch crow prints "fail" after 16 successful allocations and then remains usable (up until you try to make it `malloc` something else).

### DIFF
--- a/syscalls.c
+++ b/syscalls.c
@@ -1,5 +1,9 @@
+#include <errno.h>
 #include <stm32f7xx.h>
 #include "ll/debug_usart.h"
+
+extern int errno;
+register char * stack_ptr __asm("sp");
 
 // TODO implement a str_buffer here to avoid overflowing the TRACE pin?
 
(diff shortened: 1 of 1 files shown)


csboling:
GNU documentation notes that "you can use ‘sbrk(0)’ to find out what the current end of the data segment is", could use this to implement a query for percentage of dynamic memory crow is using.

trentgill:
Wowowowowowow!!! Thank you so much!! Will test tomorrow & MERGE MERGE MERGE!!!

trentgill:
Merged this in along with a fix to make sure the error messages are printed to the user. Tested it and it seems to work really well from druid. you can allocate a huge table which will succeed up until the memory runs out. At this point even trying to `print'hi'` will fail with `not enough memory.` but if you clear the table, the system is back to normal!

Thanks so much @csboling !

Issue #67, opened 2019-02-20
'Composite slopes' for output library


trentgill:
Composite slopes would be composed of three elements:
- offset with a dedicated slew time
- actions with an assignable modulation added to offset
- pulse functionality overlaid on all of the above

The general use-case in mind here is an extended version of teletype commands:
```
OUT
OUT.SLEW
OUT.PULSE
OUT.SHIFT
```
So an output has a `offset`(commonly a pitch) which it moves toward at the `slew` rate, while a `mod`ulation (typically an lfo for vibrato) is applied. then atop this a triggered event can be added too. In TT the `.SLEW` parameter is used across both 'offset' and 'pulse' functionality. For a `pulse` it becomes the attack & release times. Here we allow a completely customizable `pulse` which need not be a square wave smoothed by slew, but rather a generic envelope which can be triggered by a physical input or otherwise.

For example, the `slew` section could be articulating a generative melody with portamento, while having a fast but subtle vibrato added by `mod`/`action`, with the `pulse` section instantly jumping up an octave every 4th note, then gliding back down every 8th.

All three of these functions will be implemented as ASL slopes, composed together in the IO_BlockProcess() function. This requires some small extensions in the C-layer, but primarily requires development in the lua Output library.

consider the existing syntax:
```
output[n].asl.action = {asl}
output[n].asl:action()
```
to make this clearer in the 3-layer setup we could use:
```
output[n].pulse = {asl}
output[n].mod = {asl}
output[n].slew = {asl}
output[n]:pulse(bool) -- nil or true begins the pulse, false ends it / goes to release
output[n]:mod(bool) -- restarts the modulation if nil or true, else goes to 'held' if it exists
-- slew() need not be called explicitly, it will be automatically called when the .offset param is updated
```
pulse uses the `.width` param for the pulsewidth in ms.
mod uses `.level`, `.speed`, and `.shape` params
slew uses `.offset` and `.rate` params. note that changes to `.offset` (and .rate?) will trigger a new ASL breakpoint, causing the slew effect.

there is clearly a great deal of ambiguity in the naming when it omes to the three speed-based parameters. i've tried to make them work with the traditional nomenclature, ie: `slew.rate`, `modulation.speed`, and `pulse.width` though perhaps these aren't intuitively obvious.

could prepend with `P`, `M` and `S` but this feels like a hacky solution. thoughts?

trentgill:
[...]

Issue #13, opened 2018-09-24
Synthesis


trentgill:
Fixed architecture dsp functionality.

Likely needs to be a global setting to switch the module into acting as a full polysynth.

(Triangle (ramp/waveshape) -> Filter | Noise -> Amplifier). 4-voice goal. Making a 2op FM option would be maximum suggested spec.

Smoothing on the control-rate inputs (from ADCs, or MIDI, or via USB).

LFO and Envelopes should be provided, potentially by the ASL language if that's possible. Might require something far weirder under the hood, so just start with the basics here.

Use as much from JF as will fit on the flash / in ram / in cpu time.

last feature on the sales pitch hah.

trentgill:
```lua
--- audio processors
osc{ freq = _
    , wave = _ -- a string ('sine', 'pulse' etc) or a number (-1,1) to xfade
    , symmetry = _
    , fm_source = _
    }
osc( freq, <opt. wave, symmetry, fm_source> )
-- backend is chosen based on args (re: CPU)

filter( input, freq, resonance, mode, model)
volume( input, level )
gate( input, level, mode )
input( channel )
noise()

-- controls
vactrol( level, time, symmetry )
slew( input, speed, linear/expo )

-- arithmetic
add() -- or mix()
mul() -- or gain()
```

trentgill:
Make a synthdef and apply it like an ASL:
```lua
output[1].action =
  volume( env()
        , filter( freq
                , osc( freq )))
```

trentgill:
Feels like i'm reinventing the wheel here... perhaps we could just use another lang's syntax?

Perhaps it should be fixed architecture synthesis and have a way to set the parameters from Lua such that they are applied in realtime. This would be particularly interesting if the modulation sources (LFOs/envelopes), could be ASL structures.

trentgill:
Coming back to the initial comment up top. Fixed architecture is the way to go:
2xVCO (either in parallel or 2op FM)

```
wave * level +
wave * level +
       noise +
       input + mix -> filter -> gate
```

trentgill:
Close in favour of algorithmic waveforms in ASL2

PR #303, opened 2020-04-17
Asl2

trentgill:
Restructuring ASL to allow them to exist without being tied to a hardware output channel.

The goal is to have a dynamic dsp graph where multiple ASL objects can be added together to form a hardware output, or (soon) used as modulation signals in an on-board synthesizer patch.

#13 #67


trentgill:
closing in favour of a different approach.

Issue #306, opened 2020-04-20
ii.faders.get for channels >2


desolationjones:
ii.faders.get doesn't return values from faders 3 onwards.

trentgill:
Fixed by #307

PR #307, opened 2020-04-20
add getters for faders 3 - 16

discohead:
I'm not actually at my rack to test this out. I just saw https://github.com/monome/crow/issues/306 and I'm guessing this is the fix.


trentgill:
@discohead thanks so much for that! I believe your spot on and the build system thinks it's fine too! Hopefully we'll get that confirmed with the next update.

trentgill:
fixes #306

Issue #279, opened 2019-12-23
ii communication with ER-301 is lost (until reboot) if a message is sent while the pull-ups are set to false


nordseele:
Here's a possible issue with the firmware 1.02 discovered as I was experimenting with the ER-301 via I2c from Max (Max -> Crow -(ii)> Er-301. | ER-301 is the only device on the bus, directly connected to Crow

If a message ( i.e: `ii.er301.tr_pulse(1)` )  is sent to the ER-301 while the pull-up resistors are turned off, it's impossible to re-establish the connection/ send another message to the 301 after having turned the pull-up resistors on again. A reboot or `^^reset` is required.
(please see screenshot)

I think that the pull-up resistors are "on" by default when Crow starts, so in theory there's no particular reason to turn them off and on again but it might be interesting to know this issue can happen (unless the issue is related to my setup).

1 -> ✅
2 -> ✅
3 -> ❌
 
<img width="907" alt="Capture d’écran 2019-12-23 à 17 24 05" src="https://user-images.githubusercontent.com/47527859/71369615-994fca80-25ab-11ea-9b0e-15de4ea2e446.png">

Can you reproduce this issue ? 

[details="here's a patcher"]
<pre><code>
----------begin_max5_patcher----------
4824.3oc6c01aiibj9yd9Uv039PNbdE5WX2M4g6BtraRlC3lMWvtygEGlDXP
I0VlioH0QR4wdCR9se8KjTjT7klVMkrSl4CdHaxVrpmtppqtphM+Ku6pqWl7
DO6Zm+UmO4b0U+k2c0UpljMbUw4Wc81fmVEEjotsqWkrcKON+5azWKm+Ttp8
uOINluJ2IOw46SS9hy29qKuk38aCii34p9CKZLbspWIK+72R.k2YV9yQbU60
5ax97xNCJZcWP9p6Ci2bap3IpIdj6BvMNtH4egH5BxMNHvBfyeV1i+56dm7O
2bhrHdPN5tj37rveQQ+P5BvwLpK0BLJAn3TDCH+Or9DWKyonSkSI1XHE5O+b
J7T4TWqI7ZGFcKOKKXC+HFMLbwt8QQ628qtKHJi+OeiinEdJF.Wjmdq3RY7e
ET2Zw8kmtuuaqSPC0EnAq.MIkrJIJIUyyfE99T.y6lNOBJ49NP6JAS4CINXq
[...]

PR #327, opened 2020-06-19
ii error reporting improved. recovery from bad state

### LINKED ISSUE #308: If i2c pullups are enabled, don't suggest enabling them
When the i2c communication lines are detected as being in the 'busy' (low) state for long enough that a transmission fails, crow tells the user to try enabling pullups.

If the user has already enabled pullups, but is still getting this error, the message should change to something like "bus error". Is there anything more descriptive helpful we can tell people? Perhaps pointing folks at the i2c user guide, and a specific section that talks about insufficient pullup current?

CC @dndrks @tehn @csboling

trentgill:
Fixes #308 

This is a bigger change than expected.

I added some infrastructure for passing an error callback to the hardware i2c driver (ll/i2c.c), so we can handle error reporting form the application ii layer (lib/ii.c). Added state management to know if the pullups are currently enabled/disabled in the low level driver.

As ii leader:
Previously sending a command (Tx) while the bus had a reverse connected / frozen device, would be fine after the issue was resolved. On the contrary, a getter command (TxRx) that failed due to a bus error would leave the i2c driver in a bad state that couldn't be escaped even after the issue was resolved. This has been remedied.

^^ I don't have an ER301 to test, but this could very likely solve #279 if @nordseele wants to check out this PR.

Either way, it's ready to merge as it improves the standard failure behaviour in the known contexts.

### DIFF
--- a/lib/ii.c
+++ b/lib/ii.c
@@ -37,6 +37,7 @@ typedef struct{
 static void lead_callback( uint8_t address, uint8_t command, uint8_t* rx_data );
 static int follow_request( uint8_t* pdata );
 static int follow_action( uint8_t* pdata );
+static void error_action( int error_code );
 
 static uint8_t type_size( ii_Type_t t );
 static float decode( uint8_t* data, ii_Type_t type );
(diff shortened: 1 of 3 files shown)


nordseele:
Sure, I can test this tomorrow morning!

PR #345, opened 2020-07-06
crow.reset now returns all input & ii events to their defaults

trentgill:
Fixes the norns bug where setting the clock source to be crow would cause 'First' to restart when changing script.

`crow.reset()` is now much closer to a full reset (without trashing the lua VM or dropping the usb connection).

@dndrks @tehn if one of you could test this no longer repro's the bug that'd be great, but i'm 99% sure it's fixed


tehn:
looks good. will test!

PR #352, opened 2020-07-23
clocks running with internal clocksource

trentgill:
DO NOT MERGE
going to combine this with a number of other big features for v3.0. also this is just the beginning of support which otherwise needs much testing and improvement.

//

stayed up late and made the norns clocks work on crow!

there are some limitations:
* timing accuracy is limited to 1ms
* timing will drag when doing a lot of processing
* no interface to set the things that are done from norns' param menu

//

so i'd love to ask for some input on how to:
* set tempo
* start/stop the internal clock
* syntax to select clocksource

also curious if anyone wants to outline specific use cases (and syntax) for using an external clock. i'm imagining:
* crow input trigger (should the jack be exclusively for this? or in parallel to normal processing?)
* clock from norns (either by capturing 'crow clock output' from a norns script, or as an additional output on norns)
* clock from midi (forthcoming)

//

and it needs testing! i think the timing might be really sloppy due to how i'm handling the timebase. this can be improved by using a hardware timer on higher priority, but that's something to deal with if there are real world problems.


trentgill:
Closing in favour of #393

PR #366, opened 2021-02-04
Fix buffer allocation when reading a userscript

### LINKED ISSUE #365: Files with trailing \r cause "out of memory" on reset
from @dndrks 
"i had a support case where a user couldn’t get their script to persist between power cycles — when crow would reboot, it’d report “not enough memory”. i was able to resolve it by removing the trialing carriage returns at the end of the file.the user was wondering why the carriage returns would cause this trouble and if it’s within druid’s scope to remove trailing spaces / new lines automatically (they offered to PR a fix)?"

offending file attached & zipped to be allowed by github:
[moving-quantizer.zip](https://github.com/monome/crow/files/5919864/moving-quantizer.zip)

occurred both on MacOS Big Sur + MacOS Catalina (10.15.5)

trentgill:
Fixes #365 (cc @dndrks )

Tracking down why adding an additional new-line to a script could cause it to trigger the 'out of memory' error. Turns out it had nothing to do with the line-endings, but rather was an issue caused by writing over the end of a malloc'd char array.

Essentially we were allocating an exact number of bytes to hold the userscript in memory, but were then filling it with whole-words (4 bytes) rounded-up to the nearest word. This caused 0 to 3 bytes to be written past the end of the buffer, trampling something in the heap.

I think in this specific script it was the precise length that caused "out of memory" error to occur, but the specific error that it caused was totally dependent on the size of the script (which would change where in memory the buffer would be allocated). Obviously, *most* of the time this didn't matter (because the heap is generally sparse after boot).

I have a sneaking suspicion this has also been causing the 'won't boot & need to 'erase_userscript' bug -- In that situation the system *is* booting, but the USB fails to enumerate, making it impossible to communicate with the device. I propose that the reason USB wouldn't enumerate is the Lua environment hangs because some state variable in memory is corrupted.

### DIFF
--- a/lib/flash.c
+++ b/lib/flash.c
@@ -107,17 +107,27 @@ char* Flash_read_user_scriptaddr( void )
 uint8_t Flash_read_user_script( char* buffer )
 {
     if( Flash_which_user_script() != USERSCRIPT_User ){ return 1; } // no script
-    // FIXME: need to add 1 to length for null char?
 
-    uint16_t word_length = ((*(__IO uint32_t*)USER_SCRIPT_LOCATION) >> 16) >> 2;
-    word_length++;
+    uint16_t byte_length = Flash_read_user_scriptlen();
+    uint16_t word_length = byte_length >> 2; // drops 0-3 chars
+    uint16_t trailing_bytes = byte_length & 0x3; // handle last 0-3 chars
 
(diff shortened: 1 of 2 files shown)


tehn (approved):
nice fix! messy stuff here

PR #372, opened 2021-03-15
Adc timing improvement

trentgill:
Refines the timings of the ADC driver for more accurate readings and linearity across the voltage range.

The improvement is minor, but it's certainly *better*, and the big thing here is i added a bunch of documentation explaining some of the choices.

Initially I was trying to understand why there was a weird delay required in the ADC driver, so used the logic analyzer to understand what was going on. There was a bunch of weirdness (eg wrong transmission sizes), but the core problem is that the ADC is consistently raising a `!DRDY` error which resets the internal smoothing filter, meaning every sample was being filtered from scratch (rather than a running overage). I spent a solid 5 hours trying to reconfigure the clocking setup to solve this problem, but in the end had to admit defeat.

Basically I think it requires changing either the hardware pin assignments (to use the I2S interface), or some deeper architectural changes that allows the sampling clock (a PWM channel on the stm32) to be phase-locked with the DAC's MCLK (runs on I2S PLL at 48kHz which requires a custom PLL clock unavailable to the PWM channels).

//

Realistically we can add a moving average in the DSP realm if the values are not consistent enough. Even 3 samples would be plenty, and still allow 500Hz rates.

### DIFF
--- a/ll/ads131.c
+++ b/ll/ads131.c
@@ -64,11 +64,20 @@ void ADC_Init( uint16_t bsize, uint8_t chan_count, int timer_ix )
     adc_spi.Init.CRCPolynomial     = 7;
     if(HAL_SPI_Init(&adc_spi) != HAL_OK){ printf("spi_init\n"); }
 
+
+// 216/27 = 8MHz. then /2 and /2 in the ads clocks so 2MHz fMOD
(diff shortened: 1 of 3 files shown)


tehn (approved):
thanks for tackling this!

PR #373, opened 2021-03-21
Scales can be dynamically set

trentgill:
Previously to update the output.scale setting, you would have to call the function with the full description.

Now you can just set the scale member in the output table and it will update the C layer in the background, preserving any other manipulations (like scaling & mod size).

```
output[1].scale({0,4,7,11,14,17,21},1.0,24) -- setup a 2-octave fully extended major arpeggio
output[1].scale = {0,2,4,7,9,21} -- change the scale, but maintain over settings
```

### DIFF
--- a/lib/lualink.c
+++ b/lib/lualink.c
@@ -272,6 +272,13 @@ static int _get_state( lua_State *L )
 }
 static int _set_scale( lua_State *L )
 {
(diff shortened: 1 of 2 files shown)


tehn:
dang this is cool!!

PR #374, opened 2021-04-12
fix bug where aliasing an ii address before use could stack overflow

### LINKED ISSUE #341: Crash caused by aliasing an ii device before first use
the following snippet causes a crash after a few repetitions. after some investigation it (at least once) hard faults inside the lua garbage collector. probably because the C stack is growing (from an interrupt) while the GC is running because lua used all the ram.

```lua
--- this will crash after a few seconds
s = ii.wsyn
metro[1].time = 0.5
metro[1].event = function() s.play_note( 0, 3 ) end
metro[1]:start()
```

But simply accessing the ii device before creating the alias solves the issue:

```lua
ii.wsyn.play_note(0,0) -- call something in the module to be aliased first
s = ii.wsyn
metro[1].time = 0.5
metro[1].event = function() s.play_note( 0, 3 ) end
metro[1]:start()
```

This is related to the way the ii libraries are *not* loaded into ram until a script uses them (as they take up an awful lot of ram with all the functions, and it's rare to use more than a couple at once). the issue being that `ii.wsyn` points to a different object before and after the module is loaded.

Seems like perhaps the pre-call alias is recreating the ii table for the module (or perhaps `dofile`ing the whole module again)?

Perhaps the solution is that instead of redirecting `ii.wsyn` to a new table after `dofile`ing, instead we redirect the initial pointer (`ii.is.wsyn`)?

trentgill:
Putting this here as I was staring at the code for an hour last night and think this could solve the problem in #341 

Testing looks something like:
```
s = ii.wsyn
for i=1,100 do
  print(i, collectgarbage'count')
  s.ramp(0)
end
```

Without this fix, the RAM should be exhausted within a few iterations. This fix should allow substantially more without crashing (and perhaps won't crash at all). I'm reasonably sure the above will crash on the current `main` but that should be tested too.

Super happy for anyone else to confirm this.

Fixes #341

### DIFF
--- a/util/ii_lua_help.lua
+++ b/util/ii_lua_help.lua
@@ -37,9 +37,10 @@ end
 
 function Is.openlib( self )
     local n = self.name
-    ii[n] = dofile(string.format('build/ii_%s.lua',n))
-    ii[n].help = self.help
-    self = ii[n] -- redirect the whole table (to handle extant aliases)
+    if not rawget(ii, n) then -- confirm it hasnn't been loaded (handles pre-openlib aliases)
+        ii[n] = dofile(string.format('build/ii_%s.lua',n))
+        ii[n].help = self.help
+    end
     return ii[n]
 end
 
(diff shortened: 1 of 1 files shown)


trentgill:
tested & fixed.

works well even after 100 iterations only creates ~10kB of garbage (whereas before it would cause a hard-fault out-of-memory in 23 iterations).

PR #377, opened 2021-04-13
Python based bootloader support and dfureset

trentgill:
`make pydfu` wraps the crow binary into a dfu package, then uploads using `pydfu.py` from [micropython](https://github.com/micropython/micropython) project.

Amazingly it's the fastest upload i've ever seen to a stm32 chip!

`make dfureset` attempts to send the `^^b` command to a connected crow. assumes port `/dev/ttyACM0` so it will only work on ubuntu right now.

`make dfureset pydfu` will build the firmware, reset to bootloader mode, and upload to memory.

//

@tehn the make recipe for `pydfu` is all it took to get the firmware to flash. it just relies upon `util/pydfu.py`. i'll need to update the git release to build .dfu file.

would be great if you could just try and build & upload the firmware with these new commands!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,8 +244,13 @@ debug:
 dfu: $(BIN)
 	sudo dfu-util -a 0 -s 0x08020000 -R -D $(BIN) -d ,0483:df11
 
+dfureset:
+	@stty -F /dev/ttyACM0 raw speed 115200
+	@echo '^^b' > /dev/ttyACM0
+	@sleep 1
+
 pydfu: $(TARGET).dfu $(BIN)
-	python3 util/pydfu.py --vid 0x0483 --pid 0xDF11 -u $<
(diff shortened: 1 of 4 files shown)


tehn:
whoa! you did it! yes, will test it out.

tehn:
just a forward reminder note:

```
pip install pyusb
```

(there's also a `usb` package, which is confusing)

tehn (approved):

tehn:
i don't recall exactly but maybe druid installation has dependency checking? so that install can just take care of it...

trentgill:
yes definitely druid has auto dependency handling (it already does it for the TUI library).

just checked and it looks like a one-line addition here https://github.com/monome/druid/blob/3541aab071b507c8bfc82a17deb52ec27db61bd5/setup.py#L35

PR #380, opened 2021-04-14
Just intonation support

### LINKED ISSUE #310: Input & Output 'scale' modes need just intonation support
See #14 and #292 for syntax ideas.

trentgill:
Fixes #310 

Adds general support for just intonation, as well as dedicated support inside of input & output 'scale' modes.

2 new global functions `justvolts` and `just12` for converting from fractions (ie floats) to their volts, or 12TET representation.
They have the same signature `justX( notex, offset)`. `notex` can be a single fraction which returns a converted value, or it can be a table of fractions which will return a new table with each element converted. `offset` is an optional arg which expects a single fraction acting as a just-transposition to `notex`. From my reading this seems like the way to handle key-modulation (open to other ideas here).

```lua
justvolts(1/1) -> 0.0
justvolts(3/2) -> 0.5849625
justvolts(1/4) -> -2.0 -- notes can be outside the 1/1 .. 2/1 range

just12(1/1) -> 0.0
just12(3/2) -> 7.019549
just12(3/2, 9/8) -> 9.058649 is 3/2 offset by 9/8

just12{1/1,3/2,7/4} -> {0.0, 7.019549, 9.688259} -- table call returns a new table
```
Originally I was only going to have `justvolts`, but `just12` made the input/output scalers easier, and it turns out to be way more human-readable. Exploring with `print(just12(...))` is a nice way of quickly seeing how different just ratios diverge from 12-tet. I think most crow users are probably familiar with the (0..12) semitonal representation of notes.

---

Output scale is used like so:

`output[1].scale({1/1,3/2,11/8}, 'ji')` -- a third optional arg sets volts/octave, so you can still use 1v2/octave if you want

Note the use of `ji` where you normally put the temperament. This calls `just12` under the hood, but is slightly better than doing it manually bc it enforces some default-ing behaviour.

---

Input scale is similar to output, but uses the `mode` function call (like normal input quantizers):

`input[1].mode('scale', {3/2, 9/8, 5/4}, 'ji')` -- again last arg (v/8 scaling) is optional & omitted in this example

---

All tests fine, but any suggestions on naming / calling conventions are welcome. Specifically `justvolts` and `just12`, and the `'ji'` param to scale.

### DIFF
--- a/lua/crowlib.lua
+++ b/lua/crowlib.lua
@@ -179,6 +179,30 @@ function delay(action, time, repeats)
     return d
 end
 
+--- Just Intonation helpers
+-- convert a single fraction, or table of fractions to just intonation
+-- optional 'offset' is itself a just ratio
+-- justvolts converts to volts-per-octave
+-- just12 converts to 12TET representation (for *.scale libs)
+-- just12 will convert a fraction or table of fractions into 12tet 'semitones'
+local JIVOLT = 1 / math.log(2)
+local JI12TET = 12 * JIVOLT
+local function _jiv(f) return math.log(f) * JIVOLT end
+local function _ji12(f) return math.log(f) * JI12TET end
+function justvolts(f, off) return _justint(_jiv, f, off) end
+function just12(f, off) return _justint(_ji12, f, off) end
+local function _justint(fn, f, off)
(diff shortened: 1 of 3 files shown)


tehn (approved):
i really like this design.

code looks good, haven't tested on device.

PR #382, opened 2021-04-14
Frequency detect mode for input[1]

### LINKED ISSUE #277: Frequency Tracker mode for inputs
This feature request is driven by wanting to replace my in-house calibration module with a generic (and sharable) solution.

Basic concept is there is an input mode (ie `input[1].mode('frequency')`) which counts rising-zero-crossings (with hysteresis). When the pitch has settled on a new frequency (using a fixed window averager) a callback event is called in lua, reporting the detected frequency.

This can be useful for doing musical analysis on a received audio signal, rather than just a control voltage. Perhaps some utility for working with external hardware that doesn't have a CV output, eg: capture and process the pitch of a vintage analog synth, adding harmonies.

For calibration, the first output is sending an output voltage (say 0V). This voltage is sent to the v8 input of the module under test, it's output (square wave preferable) is fed back into crow's input. After the pitch is detected, it is saved in a global, then the voltage is incremented by 1V, and the process repeats. testing can operate over n-octaves, and the received frequency can even be compensated by a static offset to ensure the tracking is performed, centered around some specific frequency. In oscillators, this means you could calibrate based on the 7-octave range 50Hz-3.2kHz, if you wanted to guarantee certain tracking in that region.

The output collected is absolute frequency at the given output voltage, resulting in N-values. These could be processed into their % difference from the expectation (based on the central frequency multiplied by the expected ratio). Overall correlation between expected & measured could be output on output[2]. High-frequency tracking could be available on output[2], displaying the relative % difference in the top 1 or 2 octaves, for devices that provide an HF tracking trimmer. output[3] could be a configurable 'in-range' output that goes high when the calibration has succeeded the tests -> plug this into a VCA over an oscillator to create an audible 'success' tone.

//

If the detection should report an average pitch of a streaming basis (not change-events), the time-window-average-size should be configurable to allow the scripter to prioritize accuracy vs latency as well as introduction of slew.

trentgill:
Closes #277 

DO NOT MERGE

This PR adds a new input mode `freq` short for `frequency tracker`. It functions the same as `stream` & `volume` modes, but returns values representing the frequency of an attached oscillation.

The detection algorithm is just looking for zero-crossings, so it will only work with raw oscillator sounds (though perhaps could be fun to explore with complex waves for harmonic-locking effects).

Working accurately up to about 1500Hz in my testing, with increasing jitter above this point.

//

I did notice that all the values were off by about 2%. After some investigation, it's not caused by the detection algorithm -- i tested sending LFOs from ASL, and they were also this same 2% off. It seems that the dsp block itself is running slow. I haven't investigated whether this is caused by variation between devices (it could be as I don't think our crystal is very accurate in an absolute sense - there is no dedicated audio crystal), or if the PLL that controls the audio-loop is incorrectly configured. Perhaps worth some investigation, but it shouldn't hold up this PR.

The good news is the frequency offset scales exponentially. To correct for it, one simply multiplies by 0.98 or so in frequency land. If converting to hertz, it just becomes an addition.

//

Before merging, i want to add a `hztovolts` function based on the same absolute freq range as just-friends / w/syn etc.
The reason i'm waiting is it can leverage the new `justvolts` function. Should just be:
```lua
function hztovolts(hz, ref)
  ref = ref or 261.63 -- optional. defaults to middle C
  return justvolts(hz/ref)
end
```

This is especially handy if you want to track a frequency input and control a module over `ii`, or even to take an input frequency and send it to an output as a voltage (so it can be doubled by another oscillator). This would be especially interesting if the output had a `.scale` applied, and in `volts` space, adding an offset becomes a direct transposition.

//

Together with the new calibration engine & public libs, we should make a norns-enabled script for volt-per-octave calibration of other modules.


tehn (approved):
looks great. impressed at the deep implementation, and seeing MIDI code made me smile

PR #387, opened 2021-04-21
Sequins.lua library

### LINKED ISSUE #359: Sequins.lua - musical structure utilities
From the Maps stream, I came up with some fun functions that allow for a more streamlined approach to building basic arpeggiators and such. They would likely be great additions to the crow (and norns) standard libraries.

Below is the example in question, particularly useful for generating arpeggios and basic sequences, but it could be expanded into slightly more of a library such that the 'scales' could be updated on the fly (at present they are captured & owned by the arp). Naming could be improved as there are likely implications beyond simple arpeggiators.

```lua
--- the userspace function call style
myskal = {3,6,9,15,18,21,-6}
myarp2 = run_arp( function(v) ii.wsyn.play_note(v,3) end -- what to do at each note of the arp
		, make_sequence( myskal, 'next' ) -- a generator to return the next arp note
		, 4/3 -- a timing in 'clock' module style
		)

-- takes an action, a generator, and a timing
function run_arp( fn, seq, sync )
  return clock.run(
    function()
      while true do
        clock.sync( sync )
        fn( seq() )
      end
    end)
end

-- generator that takes a table & a behaviour, returning a function that will give the next step
-- arg1: scale table eg{0,2,4,7,9}
-- arg2: behaviour enum:{'next','prev',rand','drunk'} -- n as next#
-- retv: function that generates the next note
function make_sequence( skal, behaviour )
  skix = 1
  return
    function() -- was 'new_note'
      if behaviour == 'next' then
        skix = skix + 1
      elseif behaviour == 'prev' then
        skix = skix - 1
      elseif behaviour == 'rand' then
        skix = math.random(#skal)
      elseif behaviour == 'drunk' then
        skix = skix + math.random(-1,1)
      end
      -- clamp to skal size
      while skix < 1 do skix = skix + #skal end
      while skix > #skal do skix = skix - #skal end
      -- scale 12TET notes to v8
      return skal[skix] /12
    end
end
```

Also, there is the possibility of using the clock library for (re)implementing ASL style constructs where the output is a control-rate destination. Particularly useful for `ii` destinations where we can't & don't need audio rate ASL generation. Instead, generating each 'frame' in lua is fine. Would be nice to provide versions of the the asllib.lua functions which are applied to a generic function at a given framerate. Could be useful for dynamic sample-rate reduction (read sample&hold) of the functions.

```lua
-- userspace call
mylfo = fn_lfo(
    function(v)
      ii.wsyn.fm_index(v)
      output[1].volts = v
    end)

function fn_lfo( fn, duration, min, max, fps )
  -- set defaults
  duration, min, max, fps = duration or 1
                          , min or 0
                          , max or 5
                          , fps or 30
  local phase = 0
  local range = max - min
  return clock.run(
    function()
      while true do
        phase = phase + 1/(duration*fps)
        while phase >= 1 do phase = phase -1 end -- wrap
        -- TODO add 'shapes' here. currently just triangle. ramp/sawtooth is even easier
        local o = phase * 2      -- triangle
        o = (o > 1) and 1-o or o -- triangle
        o = o * range + min -- maps to [min,max]
        fn( o )
        clock.sleep(1/fps)
      end
    end)
end
```

trentgill:
Fixes #359

Introduces a new core library to crow for imbuing tables of data with 'behaviours'. Particularly focused on note sequences, but flexible enough to be useful for many things (ideas: building markov chains, event sequencing, generative arrangement).

```lua
--- reference
-- data can be anything a lua table can store
words = {'hi', 'there'}
notes = {2, 9, -4, 0}
fns = {lfo, pulse, ramp, ar}
tabs = { {2,3}, {'even', 'different', 'types'}, {4,5} }

seq = sequins{words} -- *copies* table into a sequins and stores in seq
seq() --> returns the values from the sequins ('hi', 'there', 'hi', ...)

tab2 = {1, 2, sequins{3, 4} } -- creates a table where one element is a sequins
seq2 = sequins{tab2} -- note how tab2 contains nested sequins which will be resolved at runtime
seq2_ = sequins{1, 2, sequins{3, 4} } -- these nested sequins can be written inline
--> calling seq2() returns 1, 2, 3, 1, 2, 4, 1 ... -- note how the nested sequins vals are interleaved

seq = sequins{...}:step(-1) -- change increment step
seq = sequins{...}:select(2) -- set the table index which will be used for the next call
seq:select(3)() --> after selecting, can call the result to get the value immediately

--- method chains
-- these fns modify the state of the sequins destructively
sequins{fns}
  :step(-2) -- change increment value per call. indexes are wrapped to the table length
  :select(5) -- sets the index to be selected on the next call. value is wrapped into range
--- the following method chains are mainly for use inside nested sequins
-- they manipulate behaviour of the parent sequins
  :every(2) -- only returns a value every 2nd call. if no value returned, parent sequins will try to get the next value
  :times(3) -- forces the parent sequins to ask it for the next 3 elements
  :count(10) -- will only provide values the first 10 times it's requested, then returns nil.
  :all() -- like times(#sequins) where the nested sequins will return all of it's values once
  :once() -- like count(1)
```

Some examples to demonstrate use-cases:
```lua
s = sequins

--- sequencing lydian arpeggios with a wholetone run flourish
lydian = {0,2,4,6,7,9,11}
wholet = {0,2,4,6,8,10}
melody = s{ s(lydian):all():step(3)
          , s(wholet):all():every(3)
          }
metro[1].event = melody
metro[1]:start()

--- using a sequins inside a method chain modifier so behaviour changes over time
arp = s(lydian):step( s{2,2,-2} ) -- lydian scale in thirds

--- arranger can be built such that a single function can be called to get 'next note'
-- while changing sequence is done asynchronously
mel1 = s{0,7,2}
mel2 = s{3,6,9,13}:step(-1) -- runs right to left
mel3 = s{3,-3}:select(2) -- starts from the 2nd element
arr = s{mel1, mel2, mel3}:step(0) -- step(0) means we'll always stay on the given melody
metro[2].event = function() output[1].volts = arr()/12 end -- install arrangement into a metro
metro[2]:start() -- every metro tick will get a new note
-- now we can modify the sequence asychronously
arr:select(2) -- switch to the second sequence
```


trentgill:
Specifically I'd love feedback on naming & syntax. Use of `every`, `count` and `times` are all somewhat ambiguous but i'm also trying to keep them short so it's more readable.

The basic usage seems very clear & satisfying, but I'm not entirely sure I've approached the more complex ideas the best way. Having nested sequins is really powerful, I just want to make sure i'm providing the clearest interface into that complexity.

tehn:
super exciting, i'll sit with it and let it sink in so i can give proper feedback on syntax.

fyi it looks like some `public` things got into this commit, not sure if you were trying to separate them

trentgill:
thanks for the note re: public. i'd like to keep them separate for now, so i'll go do some git fu...

tehn:
i really like the syntax-- that the simple things are straightforward and nesting allows for some weird meta-exploration.

a full tutorial/study on this would help people see the full potential. looking forward to playing!

tehn (approved):

rbrt-fm:
i've been playing around with this for a bit this evening and it's super fun, but I've been getting a runtime error in druid when trying to use the select method 
`repl:1:1 attempt to call a nil value (method 'select')`

trentgill:
@rbrt-fm i just pushed a new commit fixing the select method. lmk if it fixes your problem and i'd love to hear your thoughts on using the library!

PR #388, opened 2021-04-21
MIDI support officially removed

trentgill:
I've gone through and cut out all the old MIDI infrastructure we built in hoping. The hardware is not reliably capable of handling MIDI input, so no need to keep it.

The benefits:
* reduced firmware size by ~5kB
* saves almost 1kB of RAM in lua env


tehn (approved):
it was a nice idea!

trentgill:
haha yes it was! shame i didn't spend more than a cursory glance 'yeah it's working' before we ran the first batch... still it's grown into a pretty wild little platform :)

PR #389, opened 2021-04-21
Refactoring & trimming lua libs

trentgill:
Work in progress.

I'm working through the lua libraries to trim unused code, and optimize for reduced RAM usage as we're rapidly running out of availablle user space RAM. This is exacerbated by the inclusion of multiple new lua libraries in the 3.0 update.


tehn (approved):
nice cleanup

PR #392, opened 2021-04-22
Stm32 Cube F7: update version & trim

### LINKED ISSUE #379: switch to trimmed stm32 cube
currently we pull in the giant stm32f7 cube in its entirety.
i made a trimmed version without all the examples, docs & RTOS etc which is about 10% the size.
this should dramatically speed up github actions, and makes less of a chore to get started with development.

trentgill:
Fixes #379 

Dramatically reduces size of repo by switching to a separate libs for the stm32 HAL libs.

This jumps forward from HAL v1.7 to v1.16 so there could be some breaking changes (i didn't read all the changelogs). There was some changes to the USB Device driver which I have updated here as well.

One nice benefit is the GH Action that builds each new commit will be much faster now without having to download 2GB of data for the HAL!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,9 +10,6 @@
 [submodule "submodules/lua"]
 	path = submodules/lua
 	url = https://github.com/trentgill/lua.git
-[submodule "submodules/STM32_Cube_F7"]
-	path = submodules/STM32_Cube_F7
-	url = https://github.com/trentgill/STM32_Cube_F7.git
 [submodule "submodules/STM32CubeF7_Drivers"]
 	path = submodules/STM32CubeF7_Drivers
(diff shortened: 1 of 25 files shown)


tehn (approved):
fantastic!

where would you expect to see breaking changes? (provided it compiles, are you suggesting runtime issues?)

trentgill:
i'm guessing it'd just be in the hardware i/o. i think i2c is the only hardware element not in use in my current test setup (which is just a crow). everything is probably fine, but just wanted to suggest further testing is needed before merge. or if we merge to the THREE branch, to note that this commit could be something to try rolling back if there's general weirdness.

tehn:
yeah i'd say we just merge to the THREE branch. (generally seems good to
just keep merging into it as a dev branch to ease/facilitate testing)

PR #393, opened 2021-04-22
clock.lua module added

### LINKED ISSUE #339: Norns 'clock' module support on crow (global timing)
Add norns 'clock' module support to crow.

This allows for a variety of clock synchronization functions, and introduces the notion of a global clock on crow.

While the implementation will be substantially different from the norns version (using a single hardware timer for the clock base instead), it should present the same interface to the user.

- [ ] Reserve one clock for global tempo & simply call into lua each tick (internally generated)
- [ ] Implement coroutine style clock synchronization (norns syntax)
- [ ] Add an event system for managing multiple simultaneous coroutines
- [ ] Add 'clock source' option to drive from a gate source (like 'change')
- [ ] Add 'clock source' option over serial device (norns, max, ableton)

### extended features
timeline support. ie a table of time, function pairs that will be executed. useful for 'composition' style sequencing.

trentgill:
Fixes #339 

This is a basic implementation of the the norns `clock` module. It's based on the initial implementation by @artfwo rather than the recently updated one. I will lilkely revisit this before 3.0 to refine some things. Especially using the CV input as the clock source needs some refinement (it seems quite drifty, and perhaps there are double-triggers).

Usage is similar to norns, but with some key differences & helper functions that are specific to the crow workflow:
```lua
--- All of the basics are identical to norns:
myclock = clock.run(function()
  while true do
    print'hi!'
    clock.sync(1/4) -- synchronise to the next 1/4 of a beat
  end
end)
clock.sleep(n) -- sleep for n seconds
clock.suspend() -- suspend until manual restart
clock.resume(myclock) -- resume a suspended coroutine
clock.get_beats() -- beats elapsed since starting
clock.get_beat_sec() -- convert beats to seconds at current tempo

--- Differences on crow are mostly about syntax relating to tempo & clock source
-- internal timebase (default)
clock.start() -- selects the internal timebase & starts the clock (this is effectively run at init())
clock.stop() -- selects the internal timebase & stops the clock [This doesn't actually stop though?]

-- tempo just works as a variable
clock.tempo = 160 -- set internal tempo to 160bpm (ignored if using external source)
print( clock.tempo ) -- print the current tempo (uses external source if active)

--- CV inputs as timebase
-- when choosing clock mode, internal timebase is disabled
input[1].mode('clock',1/4) -- set timebase to use input jack 1, where each trigger == clock.sync(1/4)

--- output helpers: create a tempo clock with one line
output[1]:clock(1/4) -- set output 1 to pulse every clock.sync(1/4), ie 4 times per beat
-- by default output.clock sets the ASL action to pulse(), but you can reassign it
output[1].action = ar() -- now the clock will trigger an envelope instead of a pulse
```


tehn:
this is super great! only issue:

> clock.suspend() -- suspend until manual restart
> clock.resume(myclock) -- resume a suspended coroutine

both of these will be deprecated. @artfwo and @ngwese agreed (in vid meeting recently) that they were not quite what was required for a particular use case--- and then there weren't really any sensible remaining use cases for these.

PR #394, opened 2021-04-22
Public.lua library for sharing parameters

### LINKED ISSUE #300: Public.lua - Norns-Crow distributed computation library
## Motivation
norns scripts almost universally control *crow*'s behaviour with direct commands being called by the active norns script. When a musician wants to have crow operate in a free-standing manner, they typically will write a script in a text editor & upload with the `druid` application. one goal of this issue is to unify these processes for norns users (ie, edit & upload crow-scripts from norns (via maiden)), and provide a mechanism for distributed computation on crow.

## Edit & Upload
This element simply requires a bulk-upload command to be added to the `norns.crow` library. It just extends the current functionality of uploading short code-blocks with the appropriate `upload` and `run` commands, and some timing management to make sure buffers aren't overflowed.

* Would be useful to decide on an idiom that allows crow scripts to be identified & not attempt to run them on norns. This could be in the crow file text itself, or in the naming of the file.

I have a working prototype of this, but it needs stress testing (hasn't been tested since crow 1.0.1).

# Distributed components
This section is about a mutual library that makes distributed computation on crow work seamlessly with the existing norns platform. This is useful / meaningful in two contexts:
* When crow locally transforms input (CV or ii) into an output response (CV or ii), with the transformation controlled remotely
* When a crow script is configurable remotely, but should be able to continue running even if the host becomes disconnected.

A lua library is required for both norns & crow. I'm currently calling it `sync` but this needs to be changed. These libraries will perform a handshake when connected, maintain synchronization of shared parameters, store parameters, and gracefully handle disconnection.

### norns lib
* upload a `*.lua` file from the file-system to an attached crow
* mechanism for confirming if the attached crow has already has the appropriate script running
* parameter synchronization (see below)
* call remote functions
* *save* the current parameters into crow's non-volatile memory

### crow lib
* load saved parameters on boot, with default fallback
* remote function declarations
* parameter synchronization (see below)

## Parameter synchronization
Consider that the params are 'owned' by crow, so:
* On norns this is a `remote` parameter.
* On crow this is a `public` parameter.

All actions on these parameters are just like normal lua usage, implemented with metatables, so you can assign to, and read from, a public/remote parameter regardless of which device you're on. When a param is changed it will automatically be updated on the other device (though no callback event is triggered).

Either device may declare a parameter, though the idiom shall be to declare them all in the crow init function. This is to allow crow to handle initialization from saved or default values, in lieu of an active connection to norns.

Either device may put functions into the shared table enabling the other device to call those functions.

Initially, numbers, and strings will be supported as data types. In future I'd like to add bools & flat tables, and perhaps nested tables later on.

Saving a parameter set is implemented by re-uploading the script with the list of remote params appended to the script. When crow initializes, it will check if the `sync.shared` table already has data, indicating the defaults should not be used.

## example
This doesn't do anything interesting, but it shows the way params are handled natively in both cases (just inside the `sync` namespace). Demonstrates reading & updating params on both devices, calling functions on a remote device, and load/save functionality.

```lua
--- shared.lua

local remote = crow.sync.shared -- alias for readability

function init()
    -- connect to crow, check for running script & upload if not running
    -- if crow is not connected, this will be run when the connection initializes
    crow.sync.script 'shared_crow.lua' -- idiomatically the script name + _crow
end

function enc(n,d)
    if n == 2 then
        -- updating a remote variable
        remote.length = util.clamp( remote.length + d, 1, 16 )
        redraw()
    end
end

function key(n,s)
    if n==1 then
        -- save remote variables to crow's flash
        if s==1 then crow.sync.save() end
    else
        if s==1 then
            -- call a remote function
            remote.new_word(n) -- key2/3 changes lower/upper casing
        else
            -- wait for key release so remote.string should be up to date
            redraw()
        end
    end
end

function redraw()
    screen.clear()
    screen.level(15)
    screen.move(0,10)
    screen.text( remote.word )
    screen.move(20,60)
    screen.text( remote.length )
    screen.update()
end

-- FIXME should this be public.redraw?
remote.redraw = redraw -- redraw can be called by remote
```

```lua
--- shared_crow.lua

local public = sync.shared -- alias for readability

function init()
    -- attempt to load a saved configuration
    if not sync.load() then
        -- otherwise use the defaults defined here
        public.length = 1
        public.word   = 'sharing'
    end
end

public.new_word = function(case)
    local s = ''
    case = (case==1) and 56 or 96
    for i=1,public.length do
        s = s .. string.char(case+math.random(26))
    end
    public.word = s
    public.redraw() -- FIXME? remote.redraw?
end
```

trentgill:
DO NOT MERGE. WIP.

This is the `public` library designed for sharing params between crow and a host device (eg. norns). Params are maintained on both sides in a distributed context. Basically it just means the value can be changed by either party, and the other will be informed (with the possibility of doing an action on change).

more TBD.

Needs refactor from current state

Fixes #300


tehn:
crow and norns go to a pub...

code looks great. changed target to THREE so i could see the diffs properly. nice to see the test file with all the `asserts`!

this is really powerful new functionality--- will be interesting to auto-populate these into the norns param menu!

trentgill:
-- declare a public var 'a' and initialize it to 1.1
public.add('a', 1.1)
-- update public var 'a' to value 2.2 and update a connected host
public.a = 2.2

--- public.add syntax

-- numbers
public.add('a', 1.1) -- unconstrained number
public.add('b', 1.1, {0.4,3.0}) -- number with min/max limits
public.add('c', 2, function(v) d = v end) -- number with 'action' fn that will be called on remote update
public.add('e', 10, {4,16}, function(v) f = v end) -- number with limits and action

-- enumerated type
public.add('g','a',{'a','b','c'}) -- basic enum called 'g', initialized to 'a' from set {'a', 'b', 'c'}
public.add('h','x',{'x','y','z'}, function(v) i = v end) -- enum with action

-- tables
public.add('j', {1,2,3,4}) -- unconstrained table with 4 elements
public.j = {1,2,3} -- table can be updated (read: replaced) by assignment 
public.add('k', {1,2,3}, function(v) l = v end) -- table with action (WHAT IS v IN THE FUNCTION?)
public.add('m', {6,5,4}, {1,8}) -- table with min/max limits for elements of table
public.add('n', {1,2,3}, {1,5}, function(v) o = v end) -- table with limits and action

-- indexed table
-- call .select() over the public table to enable sequins style indexing
-- TODO syntax will probably change by implementing in terms of sequins
public.add('p', {1,2,3,4})
public.p.select(1) -- select element 1 and return it

public.add('q',{1,2,3},{1,3},function(v)r=v end) -- indexed-table with limits & action (same as table)


--- types
-- limits can be overloaded with special 'type' information
-- doesn't affect local usage, but it should affect host control
-- options are not defined on crow, and can be any string
-- standard types to be supported are:
    -- 'expo': changes increment behaviour to map exponentially over the given range (time / frequency)
    -- 'slider': display the value as a slider rather than a number
    -- 'bold': display the value in a bold way (through typeface, size, etc)
-- prefix any type with an @ to make it display only (not modifiable)
    -- '@': normal value but displays only
    -- '@slider': a slider shows the val, but can't be selected or moved
-- FIXME: @ is not available for enums.

tehn:
this looks good!

for _enumerated_ type on norns we call this OPTION, and the index is a number, the OPTIONs are the "enums". i'm wondering if it' clearer or more confusing to have an init value be called by value rather than index. ie, can you have repeat values in the table?

for the _table_ type it seems the table size can change dynamically? i agree that the _action_ is a little confusing. i guess the passed value should just be the whole table itself? also the mix/max as applying to table elements seems slightly arbitrary for some reason... ie, the limit could also be for table length rather than per-element.

but i suspect you have use cases in mind so i'm happy to go with your existing patterns

given THREE is the dev branch i figure we should just merge this!

trentgill:
@tehn 
cool, i'll update to calling it 'option' instead.

the primary use-case for option-types are to store a set of keys into a table. you could make a public option of global symbols, but it seems more likely to make an explicit collection of objects with the same kind of operation / data. i sketched out some options below just to see if i could see something better: 
```lua
-- current form
public.add('operators', 'a', {'a', 'b', 'c'})
public.operators = 'b'
public.operators --> 'a'


-- make the index public, not the option list itself
-- not useful bc the operators list itself is never exposed
    -- so the remote device can't show the strings (that's the point!)
operators = {'a', 'b', 'c'} -- must allocate the option list separately
public.add('opix', 1, {1, #operators})
public.operators = 2 -- can't refer by the desired symbol 'b'
public.operators --> 2 -- need to lookup: operators[public.operators]


-- exposed table, call by reference, resolve to value
public.add('opix', 1, {'a', 'b', 'c'}) -- expose the list, but set default by index
public.operators = 2 -- selection is by reference (index)
public.operators --> 'b' -- must resolve to a value
-- it seems strange to resolve to a value, when we set by reference
-- if it were to resolve to a reference, it would *require* the option list be saved in another variable


-- exposed table, but call by reference, resolve to reference
operators = {'a', 'b', 'c'}
public.add('opix', 1, operators) -- expose the list
public.operators = 2 -- selection is by reference (index)
public.operators --> 2 -- then to get the value we index our local table: operators[public.operators]
```
i think the current syntax is best. there is nothing inherently numerical or sequential about the option list, so it feels best to avoid needing to index into it. i guess i'm thinking of option type as a 'set' rather than an 'array' - it's just a bundle of options to choose from.

i think duplicate identifiers should not be allowed (like a C enum). it's a list of possibilities, not a sequence.

---

i agree the table type is confusing & not entirely clear what it's purpose is. the behaviour is too similar to sequins, but isn't quite that whole thing. there's 2 primary things tables are useful for:

1. defining a set of notes for usage in `input.scale` or `output.scale`
2. defining a set of notes for a sequence, with an implicit 'index' for visualizing/modifying active step

in both cases (but especially 1.) being able to change the number of elements remotely seems desirable.

the second option is the same as the first with the addition of an `index` key in the table for the active-step details. it's basically 'sequins-lite'. i would like to make it use sequins under the hood, but only support the `:step` and `:select` methods, and non-nested values.

---

i'm thinking we should split the 'table' and 'indexed table' into 2 separate concepts, where 'indexed table' becomes 'sequins'.
so 'sequins' is just a supported datatype (with restrictions).

```lua
public.add('myseq', sequins{0,0,0,3}) -- public.add can make a local sequins publicly shared
public.myseq = sequins{1,3,5,6} -- assignment requires a whole new sequins
public.myseq[3] = 2 -- set element 3 to value 2 just like a normal table
public.myseq --> this is the sequins itself
public.myseq() --> gets the next value from the sequins
public.myseq:step(3) -- can treat the public var as a normal sequins

public.add('mytab', {0,2,3,1,2,6}) -- make a public table
public.mytab = {0,2,4,7,6} -- mytab is just a regular table, so assignment works like any other table
public.mytab[2] = -1 -- indexing works like normal too
public.mytab[3] --> resolves to 4, like a normal table
```

---

i take the point about min/max seeming ambiguous with tables. the 'limits' are operating on the nested elements, not the value itself (the table). that said, i think it's far more *useful* to have the limits on the values, so perhaps this is just a matter of documentation & examples?

changing the length of a table should be rare, and when it does happen, the whole table (or sequins) is overwritten. perhaps this feels like it doesn't relate to the min/max?

stepping back, the fundamental reason for having min/max on a table, is if you're using it to setup a scale. if you're configuring a quantizer (seems common), you want to have a values between (0..12) and it feels important to be able to express that in the public-param's description.

i realize this isn't really an argument for it, just expressing the use-case. perhaps you can see a different way to express it, or nest it?

tehn:
your explanations and rationale are excellent and totally clarified the use cases. i think we should keep it as is!

trentgill:
ok great. i need to do a little work to get sequins to work as the 'indexed table' type, and make sure the method calls are passed through correctly, but shouldn't be too bad.
and 'option' type too.

trentgill:
@dndrks @tehn i'd love some feedback here, on what feels more idiomatic for crow.

i've been thinking a bit about the declaration style of public params, and particularly the way that the arguments are very dynamic in their meaning. eg. the meaning of the 3rd arg 'limits' can be a table of strings (`option` type), or a `{min, max [, type]`} for numerics, and the same for tables, but the limit applies to the table vals, not the table itself. plus there is an optional `action` arg which can be in 3rd or 4th place depending on whether there are limits. basically it seems like a lot to remember.

so i'm wondering if a method-chain style like `sequins` is a better approach. also taking the initial table-declaration style from ASL2's `dyn` and `mutable` values. some initial ideas:
```lua
-- declare a public variable called 'mynumber' with the value 3.33
public.add{mynumber = 3.33}

-- declare with a number with range
public.add{mynumber = 3.33}:range{0.1, 9.99}

-- add an 'on-change' action
public.add{anothernum = 2.1}:action(function(v) print('new val! '..v) end)

-- type annotation of 'exp' for exponential scaling
public.add{frequency = 100}:type('exp')

-- type annotate to force integer values
public.add{index = 2}:range{1,8}:type('int')

-- we can add some helper fns for creating common composite types
public.add{index = 2}:irange{1,8} -- adds type('int')
public.add{frequency = 100}:xrange{10,1000} -- adds type('exp')

-- option type
public.add{choice = '+'}:option{'+','*','-','/'}

-- table type
public.add{mytab = {0,2,4,7,9}}:irange{0,12}

-- and for complex params they break up nicely over multi-lines
public.add{mytab = {0,3,6,9}}
	:irange{0,12}
	:action(function(t) for k,v in pairs(t) do print(k,v) end end)

-- sequins are similar, and limits can still be applied
public.add{myseq = sequins{1,2,3,4,3,2}}:irange{1,4}


--- reference
public.add{name=initial_value}
	-- limits, choose 0 to 1 of:
	:range{min, max} -- floating point range
	:option{elemA [, elemN ...]}
	-- type modifier (optional)
	:type('mytype')
	-- combination type-limits
	:irange{min, max} --> :range{min, max}:type'int'
	:xrange{min, max} --> :range{min, max}:type'exp'
	-- function to be called on remote update of value
	:action(function(v) ... end)
```

one big benefit here is it allows for more flexibility in terms of future updates / changes. we're not dependent on argument order and it feels less awkward than a single table initialization.

we could also remove the `.add` and instead just call use table calls over `public` to add to values (it's one of the only declared functions of `public`:
```lua
public{speed = 1/4}
```

tehn:
end to start:

- i'm pretty highly in favor of removing `.add` given it ends up being a special case whereas other public namespace happen on the same level ie `public.myvar`

- i'm astonished this works in lua:

```
-- and for complex params they break up nicely over multi-lines
public.add{mytab = {0,3,6,9}}
	:irange{0,12}
	:action(function(t) for k,v in pairs(t) do print(k,v) end end)
```

i won't lie that it feels a little weird, but mainly because i'm not used to it. i'd be good with it if you're using a similar syntax for other parts ie sequins

the less-elegant version

```
public{ name = mynum
  , init = 3.3
  , range = {1, 10}
  , type = 'exp'
  }
```

is pretty much what's used in the norns param system, though we don't need to sync the two by any means.

---

oy, was that helpful??

dndrks:
woahhhh super rad problem!!

while i agree that the syntax between norns + crow don't need to the be totally similar, this is a space where we might want to avoid introducing new idioms unless we're gong to employ them on the norns scripting side as well. i just always want folks to be able to move within the "ecosystem" with coherence -- but i totally agree that good ways to shortcut working with nested tables is a unique design challenge.

beyond parameters, i'm also reminded of [lattice](https://monome.org/docs/norns/reference/lib/lattice)'s constructors + helpers, which maybe seem like they're trying to solve a similar problem? also reminded of [Pbind](https://doc.sccode.org/Classes/Pbind.html).

but also, it _should_ feel elegant to you -- is there a way to support both the more-declarative style and the shorthand?

trentgill:
i wrote up a list of all of the syntax on crow here https://gist.github.com/trentgill/71da8b6854be5f31de8d7d837dacf5f4.

it shows all the different ways of setting & querying values, and configuring the higher-level functionality. of note, there are often multiple syntaxes for configuration (eg. setting the `input.mode` can be done in 3 different ways). i remember noting that we should have done a full review before 1.0 to determine the idiomatic options, but we didn't get it done in the rush before release.

i digress! the above is an interesting bigger topic, but i wrote it up so that we could look at the whole system from a broader perspective.

when comparing the method call style to `sequins`, i note that sequins *needs* this style, as the order of methods changes behaviour. for `public`, all configurations are commutative, so the method chain perhaps makes less sense.

@tehn i agree we should drop `.add` which brings it in step with the sequins constructor eg: `sequins{1,2,3}`.

i really like the `public{myname = myvalue}` syntax but it doesn't gel well with the full table syntax: `{name = myname, default = myvalue}`.

my primary gripe with the record-style table syntax is there's no clear indication of which keys work well with each other. secondarily, we replace having to memorize argument-list-order with record-names, eg: "was it `option` or `options`?", or `default` vs `initial` vs `init` vs init_value` vs `value`.

---

my concern with making it look even more like norns' `params` is  that we're not even close to implementing the same set of features. `public` is far more stripped down out of necessity, but also it *should* be simpler to use.

the good news is this is a 'one-time' issue to solve. once a public variable is registered, all script-access to that value is as if it's a regular variable. this breaks us out of the complexities of the `params` system, and `lattice` which constantly remind you that it's a special value. `public` instead is just a fancy variable initializer that allows you to attach metadata which will be declared to an attached host.

i think this is why i keep coming back to the `public{myname = value}` style, because it *feels* like a regular variable assignment. i accept that it's probably not the best for the above discussed reasons, but just trying to explain how i'm thinking about public variables-- they are just that, variables which are exposed to the public (host). the issue is, in order for that public exposure to be meaningful, we need to add metadata relating to what type of variable it is to be used as.

---

ok, i'm rambling here, and i think i've laid out all the points i can think of. i'd love if you both could choose your preferred style from the gist (or propose another!). we can obviously keep the existing function-args style (indeed the others would be implemented in terms of it), but it's mostly a matter of what the idiomatic style should be. ie. what does the documentation use, which fundamentally informs what scripters will use.

tehn:
the command summary is great. i'll process my preferences and post back.

dndrks:
>once a public variable is registered, all script-access to that value is as if it's a regular variable. this breaks us out of the complexities of the params system, and lattice which constantly remind you that it's a special value. public instead is just a fancy variable initializer that allows you to attach metadata which will be declared to an attached host.

ah! ok, rad -- I wasn't clear on this + seeing the full scope laid out in that gist was super edifying, thank you!

so then, yes, I totally prefer the `public{name = init_value}` approach, as it allows for easier transformations of data -- `pubvar:range{1,5}` is much cleaner than `pubvar.range.min = 1; pubvar.range.max = 5`.

and revisiting your message from a few days ago with this sense of "well, these are each methods which can be stacked", something like `public.add{index = 2}:range{1,8}:type('int')` is _much_ more legible to me now. is "stackable methods"  a good way to think about this?

trentgill:
@tehn 
No i think it's correct as is. In order for the method chaining to work, `public{name=init_value}` will return a reference to `public.name`. So you can chain directly, or via the reference:
```lua
pubvar = public{name = init_value}
pubvar:range{min, max}

-- same as
public{name = init_value}:range{min, max}

-- same as
public{name = init_value}
public.name:range{min, max}
```
I was just trying to demonstrate each function / method in isolation, which required using the symbolic `pubvar`.

@dndrks i think it's idiomatic to call it "method chaining" rather than "stackable methods", but i'm not a real OO person. I just see it as syntax sugar.

Issue #390, opened 2021-04-21
Reduce RAM footprint of core libs


trentgill:
Lua is allowed to use ~155kB of RAM before we see out-of-memory errors (and occasional crashes).

As of 2.1 we're at ~110kB of space used by the core lilbs. A lot, but still 45kB for the user script to run. With the impending lua libs of v3.0 we're rapidly approaching 155kB in the core libs themselves! Obviously this won't work, and depending on which PRs you're running, there may already be out-of-mem errors.

As a result, we need to prune back the lua libs dramatically. If possible, I would prefer to do this without removing functionality. Secondly, if we can avoid rewriting large chunks of the libs in C that would be nice too (though I'm happy to do it if necessary).

I've been working through libs and am finding inconsistent results in how to reduce the RAM footprint. Unfortunately it seems the RAM cost of functions is actually quite high, meaning we might need to inline some function calls to remove helper functions. Before doing this all over the place I would like to have an objective method by which to test whether changes are helping.

I'm thinking something along the lines of using `luac -o a.out lua/input.lua && stat a.out | grep Size` to compare bytecode sizes. I don't think the 32/64bit difference will make a substantial difference. Perhaps there are references people know of regarding the RAM usage impact of lua? We have plenty of CPU headroom, so if there's ways to reduce memory pressure with more CPU that's an option.

Would love some input on how folks would tackle this problem. @tehn @csboling @simonvanderveldt @ngwese @catfact

trentgill:
I've gone for somewhat of a bruteforce approach that strips debug info (same approach as `luac -s`) on files as they are loaded. There's obvious downsides to this, but it's a good stop gap. I'm very interested in exploring other options that are more focused.

trentgill:
fixed in #395

PR #395, opened 2021-04-22
Strip debug info - reduces RAM usage

trentgill:
This PR changes the way lua libs are loaded on crow. All debug info is stripped, so errors that happen in the core libs don't have line numbers and local var names.

I think we can justify this bc it's rare for an end-user to see stack trace errors from within the core libs. Generally errors are syntax based, or from incorrect args, most of which explicitly show error messages. It seems unlikely that a user would see an error message from a lua lib, and go find the source and track down the line where it happened.

The main challenge is when developing new lua libs. If this turns out to be an issue, it would be easy to add a flag to the libraries to state whether they should strip debug info or not.

Dramatically reduces RAM usage (110 -> 72kB for the THREE branch). Not a complete solution for #390 but certainly enough to make the `THREE` branch realistic.

### DIFF
--- a/lib/lualink.c
+++ b/lib/lualink.c
@@ -2,6 +2,7 @@
 
 #include <stdio.h>
 #include <string.h> // strcmp(), strlen()
+#include <stdlib.h>
 
 // Lua itself
 //#include "../submodules/lua/src/lua.h" // in header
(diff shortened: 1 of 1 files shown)


ngwese:
Would the change mean that a user script passing an incorrect value into a core library function would not get any illuminating information about where in the core lib something went wrong?

I would worry that this takes away end user usability.

trentgill:
@ngwese There is still some illuminating information. From the `luac` docs:

> strip  debug  information  before  writing  the  output file.  This saves some space in very large chunks, but if errors occur when running a stripped chunk, then the error messages may not contain the full information they usually do.  In particular, line numbers and names of local variables are lost.

So you should still get function names at least. I just started entering a bunch of incorrect code into the REPL, and realized that pretty much everything I could think of that was wrong either failed silently, or with a `repl:1: attempt to index a nil value (field '?')` even on the main branch.

Basically I don't have a good model of what typical scripting errors look like, so it's really hard for me to say whether the error messages will be less informative. Maybe you have some pointers on how to model typical error states?

---

Actually reading that `luac` line again, the only time we "run the stripped chunk" is when we load it (when crow starts up). Maybe I'm being pedantic though, and it just means whenever that code is called at runtime?

I will say that even when there are line-number errors, they are already different from the source files bc of how we strip the whitespace & comments at compile time.

ngwese:
Unfortunately I’m not a good data point regarding the development experience on crow. My setup isn’t conducive to it as the synth is far from a computer these days. My comment/question was largely motivated by my initial experience with crow - I found debugging harder than on norns and the error messages less descriptive. If the UX after stripping this into isn’t dramatically different than before I guess it isn’t a big deal.

tehn (approved):
after a long chat with trent with further explanation and examples i'm convinced this won't hurt the user experience at all. the RAM benefit is massive, so i'm going to merge this and keep my eyes out for any complaints about difficulty debugging.

debug info will be full/intact for the user script itself, and we concluded it was very difficult (basically impossible) to create a meaningful error that needed core lib traces--- and i'm not sure this actually helps users (often on norns the deeper trace is very confusing TBH).

PR #397, opened 2021-04-23
New calibrate.lua & underlying engine

### LINKED ISSUE #39: refine calibration engine

trentgill:
Fixes #39 (finally!)

The C interface for the calibration engine is greatly simplified, and a very simple lua wrapper is provided in `calibrate.lua`. Exposed as `cal` in the global env.

There is a new script `util/recalibrate.lua` which uses this interface to iteratively solve for exact calibration. The script prints instructions on load, and utilities are provided for viewing values `pprint()` and a flag `VERBOSE` can be set to true to watch the values converge (or fail to) when debugging. Calculated values are validated to be within normal ranges, and written to flash if successful.

Currently I've only tested on a single hardware unit, but I will test a few more shortly. The `recalibrate` script could likely be refactored somewhat, but it's functional and working well for now. Curious to see how it performs on other hardware.


tehn (approved):
dang that new recalibrate script is nice.

this is going to un-zombie a huge heap of crows!!

PR #399, opened 2021-04-28
ASL2

### LINKED ISSUE #301: Audiorate ASL support
## issue
The fundamental issue here is that after each breakpoint in an ASL we must call back into Lua. Because this happens inside the audio callback, we can't call directly into Lua as it may already be active from the event loop. Instead we cue an event that waits for Lua to process it.

The result is that there is always some delay in getting the next destination value, and thus the current audio vector just sits at the limit and waits until the next cycle (or potentially n cycles later) for the callback to have been called. We already compensate for this delay by jumping ahead the appropriate part of the waveform.

The effective result is a maximum cycle time of 2 audio-vectors (one up, one down) which makes high-quality waveforms impossible with the current system.

## possible approach

ASL could be refactored such that the ASL program is compiled into a C data structure, rather than maintained as a lua table. This makes audio-rate no problem, but adds some limitations:
  * ASL could no longer call other lua functions (ie. only the `to` function is allowed)
  * Arguments to `to` would need to be literals (not functions), so values could not be updated without recompiling the ASL.

Currently the function handling just allows params to be functions which are resolved only when applying that segment to the slope library. Thus, looping ASLs will continuously resolve these calculations allowing for new values to be applied every breakpoint.

This could be ameliorated:
  * Add a 'listener' type for `to` parameters, and a userdata table where a fixed number of variables can be stored. These would be directly accessible from C, and could be updated in realtime by the script writer as eg: `listen.frequency`. There could be a set of fixed names with special behaviour (eg 'frequency' converts Hz->S). Care needs to be taken that each channel has separate access. These params would be compared every audio-vector, though perhaps the 'at next breakpoint' behaviour of the current system would be interesting to keep?
  * As an extension of 'listener' type, there could be 'input' type which, for duration of the `to`, forwards the input to the output. Otherwise 'replacement' types (eg 'noise') could be allowed.
  * general function calls could be saved into a table in the lua asl, then called from audio callback (via the event system). This could easily blow up the system, so should be discouraged.
  * could add a set of common interactive behaviours (eg. `reset_all` would restart, ie sync, all ASL channels)

trentgill:
## A Slope Language v2

ASL is rebuilt to execute in C, allowing for audiorate cycles - specifically oscillation up to nyquist freq (24kHz).

The API has some breaking changes, but is generally the same as V1.

```lua
-- globals (same as prior)
to(voltage, time, shape)
loop(...)
held(...)
lock(...)
times(n, ...)

-- globals. newly added
dyn{k=v}
mutable(n)
mutable{k=v}

-- notable ASL namespace functions
asl._if(predicate, ...)
asl._while(predicate, ...)
```

Asllib is also updated but functionality is unchanged (The code is dramatically simplified though).

`n2v`, `negate` and `clamp` are removed (as they can be replaced with native math operations, or asl constructs. This may break some rare scripts (notably `n2v` was used frequently in non-asl code. It can be replaced with a divide-by-12).

`output[n].done` event callback is no longer available
`is_running = output[n].running` query is no longer available

---

`dyn`, short for Dynamic, is useful for naming ASL arguments so that they can be updated by a script, without restarting the ASL.

```lua
-- create an lfo with a dynamic height, starting at 3V
output[1]( loop{ to( dyn{height=3}, 1)
               , to( 0, 1)
               }

output[1].dyn.height = 1 -- update height of LFO to 1V
-- note: this will take effect at the next repetition
```

Dynamic vars can be used anywhere in the ASL, either in a `to()` call, or as a predicate value (eg. `asl._if`).

Additionally, all asllib functions may take `dyn` values as variables. Eg:
```lua
-- an lfo whose speed can be changed at runtime
output[2]( lfo( dyn{speed=0.1} ) )
```

---

Mutable values are typically literal values that can have a mutation applied at each usage.

Stepping back, both `dyn` and `mutable` values can have arithmetic applied to them. This is written as if the wrapped values were simple literals, eg `dyn{num=3} * 11 - 2`. The result will be computed at runtime, such that as the `dyn` is updated, the result will change as well.

Available operations are `+ - * / %` where `%` is floating point 'wrap' from zero to the 2nd operand. `-` may be used as unary negation, or subtraction.

Returning to `mutable` values, these values will be updated by any arithmetic operations at each usage of the value. This can be used for simple iterators (the `times` construct is implemented with a `mutable`), or more complex things like animated waveforms. There is much space to explore with this feature.

Mutable values may be named or anonymous. Anonymous are as discussed above, eg: `asl._if( mutable(10)-1, ...)`. Named mutables add named access like Dynamic vars:
```lua
-- lfo where the frequency increases cyclically
output[3]( lfo( mutable{time=1}-0.01 % 1 + 0.01 ) )

output[3].mutable.time = 0.2 -- jump to a faster FM
```

---

Notably, ASL can no longer call arbitrary lua functions. The only 'verb' function is `to()`, meaning the sole purpose of ASL is for controlling voltage slopes. This will break some older scripts, but I would suggest they'd be far more cleanly articulated as a `clock` routine. Unfortunately there's no way to provide this feature now that the runtime is entirely in C.

More commentary soon!

Fixes #301


tehn:
dang look at all of those commits!

`dyn` is fantastic, great design. looking forward to more documentation and examples generally.

(aha just found extra docs in `casl.lua`)

my feeling is that you should just merge this and we can/should test everything in parallel since THREE is basically a dev branch.

tehn (approved):

PR #403, opened 2021-05-24
fix handling of ii getters from teletype

trentgill:
Fixes crow in `ii` follower mode. It's a bit of a hack, but works well with the known usage of Teletype.

Needs testing of crow to crow, but it *should* work as before as crow triggers `follow_request` directly, rather than sequentially `_action` then `_request`.


tehn (approved):

tehn:
i'd say this is a permissible hack!

PR #408, opened 2021-06-23
ASL `done` event

trentgill:
returns the `done` user-definable event which was removed with ASL2.

this allows the scripter to trigger some event when their ASL construct has completed by redefining the `output[n].done` function. this is particularly useful now that ASL runs asynchronously from the Lua environment.

```lua
output[1].done = function()
  print 'hey the ASL is done!'
end

output[1].slew = 1
output[1].volts = 2

-- <wait 1 second>
--> "hey the ASL is done!"
```

///

previously there was also a state variable `output[n].running` which would return a boolean stating whether ASL was active on a channel. i'm unconvinced this is really useful, and also *very* easy to manage directly by using the above `done` event.

///

reference note: this PR also removes the deprecated `go_toward` and `LL_toward` which are no longer used since ASL2. additionally it optimizes stack usage inside of Casl, by switching a recursive call to a while loop.

### DIFF
--- a/lib/casl.c
+++ b/lib/casl.c
@@ -3,6 +3,8 @@
 #include <stdlib.h>
 #include <stdio.h>
 
+#include "lualink.h" // L_queue_asl_done for raising a sequence-complete event
+
 // TODO
 // add sequins data type
 // add random math operator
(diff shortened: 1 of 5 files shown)


tehn (approved):
excellentt!!!

tehn:
added to reference docs

trentgill:
just found a potential bug here.

early return at line 346 in Casl.c needs to trigger the event too. find a way to unify the control flow.

trentgill:
solved the above issue using a GOTO :see_no_evil:

PR #409, opened 2021-06-23
Fix `adsr` in asllib (and false directive handling)

trentgill:
when sending directives to ASL (eg: `output[n](false)`), a value of `false` would be treated as if there was no argument present.

As a result, typical `adsr` usage -- sending true/false for attack/release -- was broken. this PR fixes this issue.

//

@dndrks after a little feedback here, but i *think* this makes sense for adsr (and the `held` ASL construct in general), but let me know if this breaks vs previous version.

when you first assign `adsr` to an output, the `held` state is `false` meaning that simply banging the output (`output[n]()`) will seem like it does nothing. this is because it's triggering the 'release' stage, which will fade from the current state (likely 0) to 0 -- hence 'does nothing'.

my thinking is when using an ASL that includes a `held` construct, the scripter *must* be explicit about when one enters/exits the `held` sub-sequence. this means the empty call `output[1]()` will restart the ASL, but it *won't* set the held state. if `held` is set, the empty call will retrigger the attack portion & rest at sustain, but if `held` is unset it will re-trigger the release portion (likely doing nothing).

### DIFF
--- a/lua/asl.lua
+++ b/lua/asl.lua
@@ -43,11 +43,11 @@ function Asl.set_held(self, b)
     end
 end
 
--- direc(tive) can take 0/1 of false/true. ONLY has effect if there is a held{} construct
+-- direc(tive) can take 0/1 or false/true. ONLY has effect if there is a held{} construct
 -- truthy always restarts
 -- falsey means 'release'
 function Asl:action(direc)
-    if not direc then -- no arg is always 'restart'
+    if direc == nil then -- no arg is always 'restart'
         casl_action(self.id, 1)
     elseif direc == 'unlock' then casl_action(self.id, 2) -- release lock construct
     else -- set `held` dyn if it exists. call action unless no `held` and direc is falsey


dndrks:
i _think_ this makes sense, yeah.

honestly, my only practical use was only in the `^^dual` M4L device where a MIDI note-on event triggers `output[1].volts = %.4f` _and_ `output[2]( adsr(0.000,0.010,0.000,0.000,'log') )`, where the user can set attack, decay and shape. this might've actually been before the AR envelope was introduced, wild.

since i wasn't previously managing a held state, the device's historical behavior seems to be in line with what you're describing as how it'll behave moving forward -- i'll give it some test runs and adjust on the device side, if needed :)

tehn (approved):

trentgill:
@dndrks yeah i think the behaviour may have been different before (`held` would always enter, regardless of state). i think this makes more sense, and agree with you that `ar` is a better option here. `adsr` is only different from `ar` if you're managing note-on/offs.

dndrks:
rad! that makes sense, thanks @trentgill .

grabbed this branch and in testing the `ar` envelopes, i noticed that `output[2]( ar(0.00001,3) )` will not execute, but `output[2]( ar(0.0001,3) )` or ``output[2]( ar(0,3) )` will. i know that `0.00001` is a dumb val to pass in, but it was actually passed into the function as the result of rounding in a Max patch (it wasn't perfectly 0, so it pushed this limit).

should it be expected that ASL would clamp values to usable ranges, or _should_ `output[2]( ar(0.00001,3) )` execute?

trentgill:
@dndrks merging this and putting your note into a new issue

PR #412, opened 2021-06-23
Add quote.lua library

trentgill:
pulls the 'quote' library from norns (built for crow/public) into the crow firmware.

it's a powerful function global function `quote(...)` that captures any object(s) as argument and returns a string that represents that object. it's primarily useful for when you have a table, or a bunch of arguments of unknown type, and you need to send them over the serial port as a string.

this means you can serialize an object & load it in a different VM:
`reconstructed_object = load(quote(original_object))`

this is required to fix a bug in the new norns-crow-namespace PR https://github.com/monome/norns/pull/1362


tehn:
looks great, super useful

PR #414, opened 2021-06-26
Asl2 dynamic mutation syntax update

trentgill:
Changes the syntax for dynamic values in ASL2, allowing for more precise control over how parameters will be varied per-breakpoint.

The concept of `mutable` values is removed (or hidden from userspace at least), instead allowing `dyn` values to have mutation-descriptions method-chained onto them.

```lua
-- dynamic variables are same as before
dyn{time = 3} -- create a dynamic variable

-- mutation-chains can be added to change the dyn value
dyn{time = 3}:step(1) -- step will add or subtract it's argument on each usage
  -- each time dyn.time is accessed, the value will increase by 1

-- multiplication is also available for exponential changes
dyn{freq = 1}:mul(2) -- freq will double every time dyn.freq is used

-- multiple mutations can be chained for complex effects
-- these results can be wrapped into a range, for cycling effects
dyn{volts = 0}:step(1):wrap(-5,5)
  -- step up in increments of 1V, wrapping into the (-5 .. +5) range

-- parameters to the mutations can also by dynamic variables
dyn{x = 1}:mul(dyn{growth = 0.99}) -- value of 1 decreasing exponentially toward 0

-- eg: apply inside an ASL and demonstrate dynamic updates
-- the dynamic chain updates the voltage the ramp-lfo will rise to
output[1](
  loop{ to( dyn{height = 1}:mul( dyn{growth = 0.99}), 0.1 )
      , to(0, 0)
      })

output[1].dyn.height = 2 -- force the lfo height to 2V
output[1].dyn.growth = 1.1 -- increase height by 10% at each call

-- note that wrap can be used without a step or mul
-- this is useful for avoiding invalid values when setting via `dyn.*`
output[1]( loop{ to( 5, dyn{rate=1}:wrap(0.1, 10) )
               , to( 0, 0)
               }
output[1].dyn.rate = 2 -- updates LFO rate to 2seconds
output[1].dyn.rate = 11 -- wraps LFO rate to 1second
```

This last concept will make more sense when other range-behaviours are ready. I'll be adding `clamp` to limit to fixed range without the roll-over of wrap. And also adding `bounce` which will invert the operation step/mul operation when reaching the wrap point -- allowing for triangular modulations.

### DIFF
--- a/lib/casl.c
+++ b/lib/casl.c
@@ -261,6 +261,7 @@ static void capture_elem( Casl* self, Elem* e, lua_State* L, int ix )
                 case '*': allocating_capture(self, e, L, ElemT_Mul, 2); break;
                 case '/': allocating_capture(self, e, L, ElemT_Div, 2); break;
                 case '%': allocating_capture(self, e, L, ElemT_Mod, 2); break;
+                case '#': allocating_capture(self, e, L, ElemT_Mutate, 1); break;
 
                 default:
                     printf("ERROR composite To char '%c'not found\n",index);
@@ -425,6 +426,13 @@ static ElemO _resolve( Casl* self, Elem* e )
             float wrap = RESOLVE_VAR(self,e,1); // 0.1
             float mul = floorf(val/wrap); // -0.01 -> -1
             return (ElemO){val - (wrap * mul)};} // -0.001 - (0.1 * -1) => 0.099
+        case ElemT_Mutate:{
+            ElemO mutated = (ElemO){RESOLVE_VAR(self,e,0)};
+            if(resolving_mutable < DYN_COUNT){
+                self->dynamics[resolving_mutable].obj = mutated; // update value
+                resolving_mutable = DYN_COUNT; // mutation resolved!
+            }
+            return mutated;} // return the resultant value
         default: return e->obj;
     }
 }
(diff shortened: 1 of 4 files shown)


tehn:
100% this! the functionality and syntax is clear and super useful. in my playing today i was looking for exactly this sort of solution.

quick question/fr:

should the built-in asl's have dyn's by default?

ie
```
output[1].action = lfo(1)
_ = output[1].dyn.rate -- = 1
```

haven't tested yet, will try it out

trentgill:
i think built-in dyn's would be nice, but part of me wants to avoid it because it's another set of words one has to remember to control them. i've always thought off `asllib` as more of a basic demonstration of what ASL can do, but never a 'complete set' of ASL actions.

i don't think it's too much to ask people to be explicit about their usage, eg:
```lua
output[1].action = oscillate( dyn{freq=440} )
output[1].dyn.freq = 220
```
i think having the explicit `dyn` makes the script more self-documenting & requires less documentation reference.

that said! it would be fun to try it out! we should make a branch with built-in dyn's everywhere and see how it feels / if it creates any issue when wanting to use a mutation etc.

tehn:
i actually think your point is very sensible--- defining them explicitly makes everything more legible, less magic vars

PR #416, opened 2021-06-27
Delay with clock

### LINKED ISSUE #413: delay can conflict with metro
at present, `delay` is implemented in terms of `metro`. it allocates a metro, then de-allocates itself once the event occurs.

this causes issues if a script uses the `metro[n]` method of directly accessing metros. this is because an indexed metro call isn't "allocated", and so `delay` doesn't know not to use that particular instance.

rather than trying to fix this in the metro lib (which i'm skeptical is possible), i propose re-implementing `delay` in terms of `clock`:

```lua
-- untested!
function delay(action, time, repeats)
	local r = repeats or 0
	return clock.run(function()
			for i=1,1+r do
				clock.sleep(time)
				action(i)
			end
		end)
end
```

an added benefit is it enables up to 100 simultaneous delays, as opposed to being limited by metro's 8 instances.

this would technically be a breaking change because the `delay` function currently returns a `metro` reference, while now it returns a `clock` coro_id. i think this usage was very rare (most uses just discard the return value), but just calling it out as a potential issue.

trentgill:
fixes #413 

tested & works the same!

### DIFF
--- a/lua/crowlib.lua
+++ b/lua/crowlib.lua
@@ -131,20 +131,14 @@ end
 
 
 --- Delay execution of a function
--- dynamically assigns metros (clashes with indexed metro syntax)
 function delay(action, time, repeats)
     local r = repeats or 0
-    local d = {}
-    function devent(c)
-        action(c) -- make the action aware of current iteration
-        if c > r then
-            metro.free(d.id)
-            d = nil
-        end
-    end
-    d = metro.init(devent, time)
-    if d then d:start() end
-    return d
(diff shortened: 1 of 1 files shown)


tehn (approved):
yes!

trentgill:
not sure what's going on, but somehow this is breaking a script on working on. holding off on merging til i can figure out if the issue is in this fn, or in the clocks system more broadly.

trentgill:
This is ready to merge as of the fix in #422

PR #422, opened 2021-07-02
fix dofile when called from a coroutine

### LINKED ISSUE #419: Running scripts in 3.0 can break ii.wsyn
This seems like a memory issue in the system. It appears as though running a script (perhaps only a script that includes clocks?) from druid causes `ii.wsyn` to evaluate to `nil` after the script initializes.

Areas to investigate:
* Memory allocator for `^^s` in repl.c
* Clock memory allocation
* Dynamic `ii` library loader

First try and determine what exactly causes the breakage (ie clocks? ii usage? script length?)

trentgill:
Fixes #419 

When calling an `ii` function for the first time from within a `clock` coroutine, it would load the file but fail to return the table to the coroutine, abandoning it to the GC.

This was caused by the custom `dofile` implementation using a global pointer to the `lua_State`, rather than the received pointer from the VM. This breaks when called from a coroutine because a lua coroutine is essentially a `lua_State` object wrapped in a 'LUA_TTHREAD' object. Thus we would push the loaded table into the main thread's stack, rather than the coroutine's stack.

---

Additionally, a new 'strip_debug' flag is added per lua-library in `lib/lualink.c`. If you are debugging an issue in any of the on-board lua libraries, set the flag to `false` for that file so you get accurate line-numbers when errors arise.

### DIFF
--- a/lib/lualink.c
+++ b/lib/lualink.c
@@ -50,20 +50,20 @@
 #define WATCHDOG_COUNT     2        // how many watchdogs before 'frozen'
 
 const struct lua_lib_locator Lua_libs[] =
-    { { "lua_crowlib"   , lua_crowlib   }
-    , { "lua_asl"       , lua_asl       }
-    , { "lua_asllib"    , lua_asllib    }
-    , { "lua_clock"     , lua_clock     }
-    , { "lua_metro"     , lua_metro     }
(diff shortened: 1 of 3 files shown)


tehn (approved):
ooof, nice fine

PR #427, opened 2021-07-08
Fixes docker run command for bash shell in dev docs

jonpliske:
This change updates the documentation to reflect that the `ENTRYPOINT` needs to be explicitly set to `bash`, otherwise the container execution will interpret "bash" as an argument.

I believe this could also be resolved via changes to the Dockerfile: changing `ENTRYPOINT` to `CMD` will allow the behavior you describe in the doc. I'm not sure of the exact use case (i.e. where else this container is run...), but [if you're trying to set the _default_ but overridable execution flow](https://docs.docker.com/engine/reference/builder/#cmd) then `CMD` might be a better fit.

### DIFF
--- a/readme-development.md
+++ b/readme-development.md
@@ -34,7 +34,7 @@ container.
 git clone --recursive --config core.autocrlf=input https://github.com/monome/crow
 cd crow
 docker build . -t crow-dev
-docker run --rm -it -v "$(pwd)":/target crow-dev bash
+docker run --rm -it -v "$(pwd)":/target --entrypoint /bin/bash crow-dev
 ```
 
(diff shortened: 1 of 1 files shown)


tehn:
seems good to me? i don't run docker though--- @trentgill ?

trentgill:
Similarly i’ve never used the docker image. Unsure of the github actions use it, or if that’s separate?

CC @simonvanderveldt @csboling who may have some insight / be able to confirm.

simonvanderveldt:
I don't know if any other commands than what's currently defined in the entrypoint (which is make) are to be run? Personally I can't think of any so I'd just remove all arguments in the readme so it just runs the container and will use the entrypoint as defined in the image.

Issue #430, opened 2021-08-03
ii.set_address has no effect


beels:
Following up on a comment by Justmat on //// (https://llllllll.co/t/crow-v3-0-1/46585/6) and then my own spelunking in the source code (https://llllllll.co/t/crow-v3-0-1/46585/7?u=3-foot-1) it looks like there may be no way for a user to change a crow's I2C address.

`ii.set_address` sets the address that is submitted to `ii_find_command`, but it appears that the filtering of messages on the I2C bus is done at the HAL level, so the address set by `ii.set_address` isn't actually used to differentiate between individual crows.

I think the only place where the HAL layer's idea of address is set is in the call to `ii_init(II_CROW)` in `main` (obviously hard-coding the CROW1 identity, and happening before the REPL is initialized so there is no way for it to reference a value set by the user).  Can `I2C_Init` be safely called from `I2C_SetAddress` so that a user can set the address by putting a call to `ii.set_address` inside their `init` script?

trentgill:
also note, the value is not saved to flash as it should be.

PR #431, opened 2021-08-03
Re-init I2C HAL layer when setting address

beels:
Proposed fix for #430.

Confirmed to enable Teletype `CROW.SEL 2` and `CROW2: ...` when `ii.set_address(2)` is run on crow.  But is this the right way to solve the problem?

### DIFF
--- a/lib/ii.c
+++ b/lib/ii.c
@@ -115,6 +115,8 @@ void ii_set_address( uint8_t index ){
         default: break;
     }
     I2C_SetAddress( i2c );
+    ii_deinit();
+    ii_init(i2c);
 }
 
 


trentgill:
i think this should be solved in `ll/i2c.c` because it's a driver layer issue (the lib folder should be hardware agnostic). i have a suspicion there is some less-intense way to make this work than a full reset of the i2c layer as i was convinced i had working previously...

beels:
A lot to unpack here!

First off, as a practical matter I haven't been able to find a way to make the change of address take effect short of invoking `I2C_Init` either directly or indirectly.  I did some experiments with replicating what looked like relevant parts of `I2C_Init` (see [beels:experiments/set-i2c-address](https://github.com/beels/crow/compare/experiments/set-i2c-address)) but the obvious approaches didn't work and I'm really just flailing around here.  Considering also the note [here](https://github.com/trentgill/STM32CubeF7_Drivers/blob/ba1a85d22d23780634bc3f3762a39c427ffce558/STM32F7xx_HAL_Driver/Src/stm32f7xx_hal_i2c.c#L42), it seems like init of some sort is the proper way to get the driver to accept a configuration, including address.

If you want to point me in the direction of an alternative, I'll follow your lead, but at the moment I'm out of ideas.

Now, as to the question of `lib` vs. `ll`, I think it's a question of what kind of interface you want to present for controlling the address.  

The pull request as presented was intended pretty narrowly as a way to make the already exposed `ii.set_address(n)` lua funciton act as an effective runtime control of the address.  You can change the address on the fly in the middle of a patch, and you get persistence only if you include the directive in an uploaded user script.  This can be seen as either a defect or a feature, but it placed the change pretty clearly at the interface level, hence `lib`.  (I'll also note that making the change in `ii_set_address` makes the change itself very simple because the `lib` layer can state explicitly what handlers it wants to install during re-initialization.  Pushed down to `I2C_SetAddress` it becomes a bit more complex: we have to re-submit the cached pointers to the handlers we were handed the first time we were initialized.  This means we have to be careful to make sure that `I2C_SetAddress` is never called before `I2C_Init` because those pointers would be unpopulated in that case.  You can see how this works out in [this](https://github.com/beels/crow/commit/a308d38c5cb7c9b34d09a65d8072f4104431b1b3) commit from my experiments branch.)

I'm guessing that a solution in `ll` would be something in the spirit of `cal`: defined at the `ll` layer and directly exported to lua without an intervening implementation at the `lib` layer.  Such a function wouldn't be part of the regular crow scripting language and wouldn't really have a place in a user script.  As such, it probably shouldn't be exposed as part of the `ii` object that is implemented in `lib/ii.c`.  You would invoke, say, `iiconfig.set_address(3)` from druid and crow would semi-permanently change its identity.  `ii.set_address` and `ii.get_address` would be deleted.  the call to `ii_init` in `main.c` would get its address argument from flash instead of from a constant.  Is that what you had in mind?

If you are interested in something similar to `cal` with its own direct persistence to flash, I'd be happy to try to do it, but we would need to settle on the interface first.  I would also need some guidance on the flash part, since I'm not a hardware guy and have no idea where to start when it comes to claiming resources in an embedded system.

trentgill:
sorry it's been a million years since you were working on this.

i finally had a moment to look into it and i think your "experiments" branch where you shifted the logic into the `ll` file is all i was looking for. my position was just that needing to reset the I2C peripheral to have an address take effect is a quirk of the peripheral, so the `lib` file shouldn't need to know about it. it also makes the ll I2C library more reusable across different projects in case of a different `lib` frontend.

i note your point about the `i2c` object being potentially uninitialized, so i just threw a simple condition in to avoid it. i think it is preferable & idiomatic to require an explicit call to `init`, so it would be confusing to setup the peripheral by changing the address. i made the change on [this branch](https://github.com/monome/crow/blob/i2c_address_ll/ll/i2c.c#L180), but if you'd like to check in your version & extend your contributions, please do :) i'm hoping for a new release in the next week or 2.

regarding the other details, i'm not too stressed about it right now. it doesn't seem like too much of a headache to add the address change into an uploaded script. though, if you want to dig into the flash saving it'd be cool nice to have it persistent by default.

i have another branch where i extended the metatable in the lua `ii` module so the syntax will also be variable style which is  more idiomatic:
```lua
ii.address = 2 -- was ii.set_address(2)
print(ii.address) -- was print(ii.get_address())
```
i feel like that's the only real thing required to make it `cal`-ish.

//

if you do take a crack at saving the `ii` address to flash, let's open a new PR. i'll keep this one open for the moment in case you want to check in those other changes.

beels:
I'm glad to hear that we've arrived at an agreeable solution for setting the i2c address.

If I understand you correctly, you are suggesting that I can merge my experimental branch into this one to claim some credit for the call to `I2C_Init`.  I appreciate the gesture, but if you have already integrated equivalent code on your own branch, I'd say just go with what you have.

As far as flash is concerned, I'm not inclined to go down that road if setting the address from `init()` is acceptable.  I actually prefer explicit runtime setting of the address.  In my own rig I move two crows around quite a bit, and the address of the crow depends on whatever role it plays at the moment, which also determines the script that is uploaded.  So address and script travel naturally together.

trentgill:
Yes that's right - just wanted to make sure your contribution was recognized bc you did all the heavy-lifting. I'll be sure to include you in the contributors list for the release!

I agree that setting the ii address in init is preferred, and feels the most logical within the script upload workflow.

I'll close this PR now, and get the finished change merged to `main`.
Thanks again for your help!

Issue #428, opened 2021-07-20
`to(v,t,'log')` interpreted as `to(v,t,'linear')`


beels:
Effectively, "logarithmic" shapes seem to always decay to "linear" in the current version.

The issue seems to be in `casl.c:240`, where the `char *` argument to `S_str_to_shape` is extracted as a `char` on the stack.  `S_str_to_shape` expects a C-string and relies on the value of the second character in that string to distinguish between the logarithmic and linear cases.

I don't know the lua API, so can't submit a pull request, but it looks like this could be fixed with something on the order of:

```
        case LUA_TSTRING:{
            char s[2];
            s[0] = ix_char(L, ix);
            s[1] = ix_char(L, ix + 1);
            e->obj.shape = S_str_to_shape( s );
            e->type = ElemT_Shape;
            break;}
```

trentgill:
@beels wow thanks for digging all the way in and finding this! i'll get a fix pushed soon.

PR #432, opened 2021-08-03
Differentiate between linear and log

beels:
I finally got a toolchain up and running, so here is a tested solution to issue #428.

Looping through client-supplied buffer may seem a bit clunky here, but I'm trusting that the compiler will unroll the loop and inline all of the code at `-O2`.


trentgill:
I think this looks great! you're only looping through 2 characters, and the code only gets called when an ASL is compiled, so i don't see any issue with it.

Will get this rolled into the next point release!

PR #434, opened 2021-09-25
Add basic support for Disting Ex dual mode commands

ryland:
Finally had a chance to connect disting to crow, so I went ahead and added these commands from the 1.9 firmware.

Not sure if the names for dual mode should be abbreviated-- they are getting a bit long.

### DIFF
--- a/lua/ii/disting.lua
+++ b/lua/ii/disting.lua
@@ -17,25 +17,25 @@ do return
   , { name = 'reset_preset'
     , cmd = 0x42
     }
-  , { name = 'load_algorithm'
+  , { name = 'algorithm'
     , cmd = 0x44
     , docs = 'load single mode algorithm'
     , args = { 'algorithm', u8 }
     }
-  , { name = 'set_controller'
+  , { name = 'controller'
     , cmd = 0x11
     , docs = 'sets parameters via mappings'
     , args = { { 'controller', u8 }
              , { 'value', s16 }
              }
     }
-  , { name = 'set_parameter'
+  , { name = 'parameter'
     , cmd = 0x46
     , args = { { 'parameter', u8 }
              , { 'value', s16 }
              }
     }
-  , { name = 'set_scale_parameter'
+  , { name = 'scale_parameter'
     , cmd = 0x47
     , docs = 'The 0-16384 value range scales to the actual param. value range'
     , args = { { 'parameter', u8 }
@@ -107,7 +107,7 @@ do return
     , cmd = 0x55
     , docs = 'set voice velocity for note id'
     , args = { { 'note_id', u8 }
-             ,  { 'velocity', s16V }
+             , { 'velocity', s16V }
              }
     }
   , { name = 'note_off'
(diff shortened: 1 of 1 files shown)


tehn:
looks good--- don't have a disting so can't test but figure you have :)

@trentgill checking if you have opinions on name lengths

lua/ii/disting.lua
```diff
@@ -118,6 +118,56 @@ do return
...
+    , docs = 'set side/parameter to actual value'
+    , args = { { 'side_parameter', u8 }
+             , { 'value', s16 }
+             }
+    }
+  , { name = 'set_dual_scale_parameter'
+    , cmd = 0x5E
+    , docs = 'The 0-16384 value range scales to the actual param. value range'
+    , args = { { 'side_parameter', u8 }
+             , { 'value', u16 }
```
trentgill, line 133: query if we can use `s16V` type here instead? in general we try and avoid scripters having to think in arbitrary scale ranges (eg 0-16384), instead opting for a more human friendly `volts` concept which is at least in base10.

the range [0..16384] is represented by [0.0 .. 10.0] with the `s16V` type. the underlying message that is sent is identical, so it's just a matter of the representation in a crow script.

i assume the values on disting itself are represented differently than as a raw 14bit value range -- if there is a numeric readout of values, i guess it would be best to mimic those values so one's script can speak in the same units as the module itself.

sorry if this is hard to grok - i've always had a hard time describing the custom type support in the i2c description lang.

lua/ii/disting.lua
```diff
@@ -148,6 +198,46 @@ do return
     , retval = { 'state', u8 }
     , docs = '0-based loop index'
     }
+  , { name = 'dual_parameter'
```
trentgill, line 201: wonder if we can abbreviate `parameter` to `param`? this obviously applies to all the above elements too.

similarly, could `algorithm` be `algo`?

lua/ii/disting.lua
```diff
@@ -148,6 +198,46 @@ do return
...
+, pickle = -- send side/parameter as one byte
+--void pickle( uint8_t* address, uint8_t* data, uint8_t* byte_count );
+[[
+
+uint8_t side = data[1];  // side
+
+if (data[0] >= 0x5A &&
+    data[0] <= 0x5C) {
+  data[1] = (data[1] & 0xF0) | (data[2] & 0xF);
+  data[1] = (data[1] & 0x0F) | ((side & 0xF) << 4);
```
trentgill, line 236: i'm out of bit-manipulation practice, but couldn't this just be:
```c
data[1] = ((data[1] & 0xF) << 4) | (data[2] & 0xF);
```
then you could delete the `side` variable.

if there's something trickier going on, perhaps a short inline comment?

trentgill (commented):
all looks good to me overall. couple small inline thoughts & ideas, but it's in a good place already.

could you do a quick tab->spaces convert (there's a couple places where indentation gets off).

lua/ii/disting.lua
discohead: butting in here to note that the `s16v` values are kind of useless with disting i2c. To control disting via i2c you either use `set_parameter` and specify the literal parameter value or use `set_scale_parameter` / `set_controller` and send 0-16384. 0.0-10.0 is the arbitrary scale in this case.

lua/ii/disting.lua
```diff
@@ -118,6 +118,56 @@ do return
...
+      }
+    }
+  , { name = 'save_dual_preset'
+      , cmd = 0x64
+      , docs = 'save dual mode preset'
+      , args = { { 'side', u8 }
+        ,  { 'preset', u8 }
+      }
+    }
+  , { name = 'take_over_z'
```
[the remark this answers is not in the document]
discohead: in the disting manual the `0x65` command is described as `take-over/release Z`. it should be obvious to most disting users. The `z` knob on the disting can be overridden by MIDI or i2c.

lua/ii/disting.lua
ryland: I'm rusty on manipulation as well... hence a step based approach, but this is much more concise.

ryland:
All the comments are good/helpful. I'll clean everything up when I have time in the next day or or two. Thanks @trentgill

lua/ii/disting.lua
trentgill: ok i just looked a little at a video and maybe i get it now?

it seems like the 'parameter value' is not a unified range, but each different parameter has a different range (but the maximum range is 0 - 16384). if that's correct, then i guess there's nothing to do but force scripters to deal with the absolute range.

so correct me if i'm wrong: in *some* cases where the range is arbitrarily mapped to (0, 16384) it would be nice to use s16V, and even better in cases where the parameter is pitch, but some cases just use the numbers as indices or a smaller range, so you need to have precise integer values.

if that makes sense, then i agree s16V is not a good fit for an mutating param range like this. but please correct me if the above is not the issue.

discohead: All disting parameters can be set in one of two ways, either you can specify the actual value that you want with no scaling or mapping whatsoever or you can send a 14-bit value and let the disting map it to the corresponding value in the set of valid values for that parameter. For example, on the Poly Wavetable synth algorithm in the Disting EX parameter 57 is a "Chorus mode" parameter and the possible values are 0, 1, or 2. So you can either use i2c command `0x46` to set parameter 57 to 0, 1 or 2... or you can use i2c command `0x47` to set parameter 57 via a 14-bit value that the disting will then map to mode 0, 1 or 2. In neither case is a range of 0.0..10.0 sensible. None of the i2c commands on the disting have any concept of "volts", so 10.0 will never be treated as a maximum unless it just so happens to be for that particular parameter. It's more like a 14-bit MIDI implementation. @scanner-darkly has posted the i2c specification here: https://github.com/scanner-darkly/teletype/wiki/DISTING-EX-I2C-SPECIFICATION

ryland:
@trentgill I've cleaned this up a bit and am now handling all commands that use the packed side/param in pickle().

I removed all `set_` prefixes from command names in order to match convention, and removed `load_` where there was no corresponding `save_`. I think this helped enough that I didn't do any other abbreviations.

@discohead this works for me in some quick tests, but maybe you have time to try it out to confirm or offer any suggestions?

PR #444, opened 2021-11-10
Caw module can now stream a const char bigger than the TX buffer

### LINKED ISSUE #264: ii.<module>.help() truncated to 1024chars

trentgill:
Fixes #264 

Specifically solves usage of `ii.<module>.help()` for modules with a lot of commands (kria, jf, txo, disting etc).

Previously, only the first 1024 bytes would be sent, after which the USB TX buffer would be full & discard the rest of a message. This PR adds a queue system so that longer blocks of chars will be broken up across multiple USB frames.

Note, this solution only works with `const char*` strings as we don't need to copy the data anywhere. It cannot be applied to scripts that generate giant strings.

//

While fixing this i discovered that the code that marks the TX buffer as empty doesn't guarantee that the buffer has been fully consumed. In practice I was able to trample the last dozen characters or so, when spamming the USB port. Making a note here so it can be searched, but not opening an issue as i don't think it's pertinent to fix for crow's usage.

If someone needs to fix it, look at the new `USB_tx_is_ready()` fn in `usbd/usbd_cdc_interface.c` for how to check if the buffer should be released.

### DIFF
--- a/lib/caw.c
+++ b/lib/caw.c
@@ -8,6 +8,7 @@
 #define USB_RX_BUFFER 2048
 static char reader[USB_RX_BUFFER];
 static int16_t pReader = 0;
+static const char* queued_ptr = NULL;
 
 void Caw_Init( int timer_index )
 {
--- a/lib/caw.c
+++ b/lib/caw.c
@@ -54,11 +54,6 @@ void Caw_send_luachunk( char* text )
     );
 }
 
-static void queue_stream( const char* ptr )
-{
-    queued_ptr = ptr;
-}
-
 void Caw_stream_constchar( const char* stream )
 {
     size_t len = strlen(stream);
@@ -70,7 +65,7 @@ void Caw_stream_constchar( const char* stream )
         Caw_send_raw( (uint8_t*)stream, space ); // fill the buffer
 
         // here's the remainder that didn't fit
-        queue_stream( stream + space );
+        queued_ptr = stream + space;
     }
 }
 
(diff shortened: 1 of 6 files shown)


tehn (approved):
this is above and beyond!

PR #446, opened 2021-11-11
Sequins strings + other goodies

trentgill:
a set of new features, shortcuts & nice-to-haves in the sequins library:

a sequins of a string is converted into a sequins of a table of characters, eg:
```lua
sequins"abc" --> sequins{'a', 'b', 'c'}
```

printing a sequins, now prints a representation of the execution state & data. should be very helpful in terms of debugging and being able to introspect the current status of a sequins.
```lua
s1 = sequins{1, 2, sequins{3, 4}}
s1(); s1() -- advance 2 steps
print(s1) --> s[2]{1,2,s[1]{3,4}}
```

the `#` length operator is now meaningfully supported on sequins. it will return the count of items in the top-level table (ie nested tables are treated as length-of-1:
```lua
s2 = sequins{1, 2, sequins{3, 4}}
print(#s2) -- 3
```

add a new `:copy()` method so that sequins can be duplicated. this is useful if you want to have 2 duplicate sequins (or to make permutations) but maintain their table counters separately:
```lua
s1 = sequins{1,2,3}
s2 = s1:copy()
s1() --> 1
s2() --> 1
```

the `:settable(t)` method is extended to handle nested sequins. if your sequins has the same structure (eg. element 3 is a sequins in both old & new tables) the execution state of the old table will be carried across. this means you can replace more complex sequences without restarting any nested elements. NB: does not copy flow-modifiers yet (eg. `every`, `count` etc).

additionally, the 'table' argument to settable can now be a table *or* a sequins (to make the syntax a little more flexible):
```lua
s1 = sequins{1,2,sequins{3,4}
for i=1,5 do s1() end -- step forward 5 steps
print(s1) --> s[2]{1,2,s[2]{3,4}}

s1:settable{5,6,sequins{7,8,9}}
print(s1) --> s[2]{5,6,s[2]{7,8,9}} -- table-length can change too!

s2 = sequins{'a','b',sequins{'c','d'}
s1:settable(s2) -- switch to another set of data, but maintain state
print(s1) --> s[2]{'a','b',s[2]{'c','d'}} -- table-length can change too!
```

### DIFF
--- a/lua/sequins.lua
+++ b/lua/sequins.lua
@@ -1,7 +1,4 @@
 --- sequins
--- nestable tables with sequencing behaviours & control flow
--- TODO i think ASL can be defined in terms of a sequins...
-
 
 local S = {}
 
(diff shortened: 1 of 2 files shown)


tehn (approved):
fantastic!!

i know there are a few more updates coming, but it'd be great to maintain parity between sequins on crow and norns before the next releases

trentgill:
@tehn absolutely! i'll open a PR pulling all the new changes into norns when this all gets merged.

i'm mostly building stuff up in the PR list here so it's easier for me to track heh.

PR #448, opened 2021-11-14
sequins refactor for more predictable flow mods

trentgill:
aggressive refactor based around solving some non-determinism & confusing behaviour in the flow-modifier system.

basically now there is only 1 of each `every` / `count` (and `all`) / `times` in each sequins. nested sequins have their own set as well.

this is in contrast to original version which had a recursive stack of modifiers that could have n-instances of each. the issue was that people wanted to *change* the value of the modifiers, but attempting to do so would append an additional modifier.

furthermore, there was previously a confusing difference in execution when you stacked `:count(n):every(m)` versus `:every(m):count(n)`. this order-of-operations has been removed, and instead we have a fixed order of operation: first we process `every`, skipping non-matched elements. then process `count`, which freezes the `every` state until the `count` is complete.

so you can say "every 3 times you call this nested sequins, i will produce 4 values in a row, before yielding back to the parent sequins". meanwhile "times" is the number of actual values returned, thus ignoring 'every' and counting once per value, regardless if it's part of a 'count'.

////

note:
* `sequins.cond` and `sequins.rcond` are removed
* `sequins.once` is removed (use `sequins.times(1)`)
* `tostring(s{1}:every(3))` and similar will now stringify any active flow-modifiers

### DIFF
--- a/lua/sequins.lua
+++ b/lua/sequins.lua
@@ -98,41 +98,32 @@ local function do_step(s)
     return retval, exec
 end
 
--- a really tight run-time function
+S.flows = {
+    every = function(f,n) return (f.ix % n) ~= 0 end,
+    times = function(f,n) return f.ix > n end,
+    count = function(f,n)
+        if f.ix < n then return 'again'
+        else f.ix = 0 end
+    end
+}
+
+function do_flow(s, k)
+    local f = s.flw[k] -- check if times exists
+    if f then
+        f.ix = f.ix + 1
+        return S.flows[k](f, turtle(f.n))
+    end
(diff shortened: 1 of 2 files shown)


tehn:
looks great. this behavior is clarifying.

trentgill:
Added my test script that walks through all the features. run it from the /tests folder with `lua sequins.lua` and it'll give you assertion errors if the API changes.

did a little further refactoring of the flow modifiers to highlight their similarity (though i'm still unhappy with how the 'count' feedback is handled... next time.

PR #458, opened 2021-12-06
i2c reliability

trentgill:
Unrelated to the pullup-centric PR, this is a (WIP) set of improvements to the i2c driver & ii layer to better handle errors.

especially when a message fails to send on an otherwise ok bus (due to another device taking control) we now retry sending that message. there is no timeout or limit, and this does mean the "ii lines are low" message is no longer sent if pullups are already enabled.

__more to come__


trentgill:
Merging this on the belief that i tested it well back in 2021! I remember testing it well, and also the high-reliability mode being default should improve 90% of issues people have with their i2c bus failing. Will be especially useful with a crow+JF combination.

The whole firmware needs some good i2c testing, and i think this should help with most general use cases.

PR #463, opened 2022-04-05
query calls currently registered stream_handler, not default stream

trentgill:
@dndrks this should solve the issue in the norns `.query` request: https://github.com/monome/norns/issues/1555

issue is that the new norns handling doesn't use the old 'named' convention (eg. `^^stream`) but instead allocates a dynamic name under the hood. the change was made such that future API changes on crow wouldn't require a co-edit to the norns system (ie. it would 'just work').

### DIFF
--- a/lua/input.lua
+++ b/lua/input.lua
@@ -119,7 +119,7 @@ Input.__index = function(self, ix)
     if     ix == 'volts' then
         return Input.get_value(self)
     elseif ix == 'query' then
-        return function() _c.tell('stream',self.channel,Input.get_value(self)) end
+        return function() stream_handler(self.channel,Input.get_value(self)) end
     elseif ix == 'mode'  then
         return function(...) Input.set_mode( self, ...) end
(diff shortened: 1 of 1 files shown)


tehn (approved):

dndrks:
raaaad, thanks @trentgill !
what do you think about bundling a new crow update for this and the other rad bits you've deployed since last july?

trentgill:
i'm honestly not sure if it's working correctly atm. i merged a bunch of changes that i think may have introduced some instability in the system (and more liklihood for crashes after using a large script for a while). if folks have some time to run some tests & general usage then i'd be confident pushing it out -- i just don't have the cycles right now to handle a situation where the update causes issues for a bunch of people.

in the meantime, this particular issue can be solved by adding the following lines to a norns script:
```lua
crow.send("input[1].query = function() stream_handler(1, input[1].volts) end")
crow.send("input[2].query = function() stream_handler(2, input[2].volts) end")
```
(untested).

PR #471, opened 2023-02-21
Ceify metro

trentgill:
Continuing the ceify journey.

this replaces the lua implementation of the `metro` lib with a C based one. there's still a little lua stub file which does the linking because i just wanted to get it in and tested. could save a little more RAM by doing the assignments in C, but this feels like a good place to leave it and move on to bigger fish.

increases the FLASH usage by 60bytes
decreases runtime RAM usage by 2.85kB (current runtime is 90.6kB at launch)

not a lot of a benefit really (especially bc it took me 6+ hours though most of that was airplane brain), but it is a solid 6% relative to the logical minimum (about 45kB).

//

i'm realizing the big benefit to come from these routines is not so much about shrinking the amount of lua code that is loaded into the runtime, but rather about reducing the amount of dynamic allocation happening as a result of using the libraries.

going forward, i'm going to focus on libraries that allocate lots of data when you use them (sequins, clock, asl). i think `clock` is the biggest culprit as creating and pause/resuming a bunch of coroutines is i think quite heavy on the runtime.

come to think of it, i should do some analysis of what is causing the ram usage. forthcoming...

//

but for now, this feels like a solid inclusion & all of the documented usage in the reference is working correctly in my testing.


tehn (approved):
code looks good, haven't tested in hardware.

this is a great endeavor--- weirdly inspiring? i'd be curious to see some metrics about the impact on execution time (not asking you to do this!)

fantastic work.

PR #472, opened 2023-02-22
fix samplerate being halfspeed

### LINKED ISSUE #453: ASL is running at half speed
This was a bug introduced in 3.0

Should be a simple fix, but i'm more worried why it's happening. Need to validate 48kHz sample rate etc.

trentgill:
Fixes #453 

I'm not sure why this has happened which gives me pause about the solution. What I'm doing now is just increasing the frequency the I2S driver runs at by 2x. Measured samplerate at approximately 48kHz on the scope, and slope times are now accurate compared to the samplerate.

I tried to increase the incoming PLL so the constants were correct in the library, but it seems the PLL is close to maxed out already. Uncertainty why, but this still solves the problem. I've checked that Metro, Clock and Input.stream modes are all running at the correct speeds, so this issue seems to be isolated to just the DAC.

I did a little benchmarking along the way, and discovered that when all 4 channels are running an ASL oscillator, the overhead is 30~35% of available CPU power (oscs at 1k,2k,3k,4k). I was worried it would be worse with the increase of 2x speed, but this seems totally reasonable to me -- It's the primary high-priority CPU task, and leaves plenty of room for the rest of the system.

### DIFF
--- a/ll/dac8565.c
+++ b/ll/dac8565.c
@@ -40,9 +40,9 @@ void DAC_Init( uint16_t bsize, uint8_t chan_count )
     dac_i2s.Init.Standard     = I2S_STANDARD_PCM_SHORT;
     dac_i2s.Init.DataFormat   = I2S_DATAFORMAT_24B;
     dac_i2s.Init.MCLKOutput   = I2S_MCLKOUTPUT_ENABLE;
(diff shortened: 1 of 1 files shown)


tehn (approved):
nice find. agreed that the cpu usage is ok.

PR #475, opened 2023-02-22
v1.1 hardware updates

### LINKED ISSUE #474: v1.1 hardware changes
for discussion of implementation of the 2 small changes for the hardware in 2023.

## hardware-pullups for i2c bus
current interface from the reference:
```lua
ii.pullup( state ) -- turns on (true) or off (false) the hardware i2c pullups. on by default
```
original hardware has ~40k resistance applied ("weak" hereafter)
new hardware has option for ~3k3 resistance ("strong" hereafter)

2 options that i see:
1) maintain same API, but new hardware always uses the hardware pullups instead of the uC internal resistors
2) add support for a string argument to `ii.pullup` where we could use `"weak"`, `"strong"`, or `"off"`.

if we choose the 2nd option, should the `true` behaviour be the same as original hardware, or switch to the hardware pullups? and also, should the default be `"weak"` or `"strong"`?

## status LED for debugging

i think @dndrks had some thoughts on what we're trying to communicate. i believe our intention is to have this LED to help us debug issues people experience in the field where they aren't sure if the system is booting up at all.

the led is a single colour, but we can a) flash it, or b) dim it with PWM.

things that we could communicate:
* is microcontroller is running?
* is a USB device connected?
* is any part of the code in an error state? specifically DAC, ADC, USB
* have we entered a Hard Fault state, ie. crow has crashed? (i'm not certain we can reliably do this, but i can try). This would need to be solid on/off, as the uC stops executing.
* (the former may be the same as) has the userscript failed to load (i actually don't think this happens anymore)?

what else do we want to communicate? what should the dim / blink pattern be for each state?

NOTE: if we want to indicate bootloader mode separately, that must be changed in the bootloader before flashing the hardware in-house.

---

to re-iterate, we need to focus in on the primary issues that result in support inquiries. the status should be the minimal set that helps us a) diagnose the hardware as faulty, or b) help diagnose a bad cable / power supply etc. or error in a userscript.

trentgill:
fixes #474 

I2C pullups should be running now. same syntax as before: `ii.pullup(true/false)`.

For testing on hardware, please measure the resistance from each i2c pin to the +3v3 test point in both on/off states. should be under 3k3 for on, and more like >100k for off. Not entirely sure the resistance measurement will while it's in circuit though... If that's the case, then you'll need a scope. Hook it up to a Just Friends & send `ii.jf.play_note(0,1)` in a fast clock, and probe the i2c pins. When on, the digital signal should be dramatically more square than when off (i'm not even sure it'll work in off mode).

---

status_led update incoming.


trentgill:
This should be ready to test now. See above for testing the i2c pullups, and the status LED should now be working as per the discussion in #474

ll/status_led.c
```diff
@@ -0,0 +1,40 @@
...
+
+void status_led_fast(LED_SPEED is_fast){
+    fast_blink = is_fast;
+}
+
+void status_led_set(uint8_t is_on){
+    HAL_GPIO_WritePin(GPIOA, GPIO_PIN_15, is_on);
+}
+
+void status_led_xor(void){
```
tehn, line 38: could `inline` to optimize, hard to know if the compiler is smart and does it already for us

same with the other one-liner funcs

trentgill: i'll check the disassembly

trentgill: it's not being inlined, but i think that's an artifact of the HAL. honestly i'm not concerned about the speed here. there is so many other far heavier things happening at the same time.

main.c
```diff
@@ -67,7 +76,12 @@ int main(void)
             default: break; // 'C_none' does nothing
         }
         Random_Update();
-        clock_update();
+        uint32_t time_now = HAL_GetTick(); // for running a 1ms-interval tick
```
tehn, line 79: would it make sense to just run this on a hardware timer at 1ms? or did we run out of hw timers?

tehn (approved):
looks great!

ll/status_led.c
tehn: yeah it's minor, no problem. cdc isn't the same as audio rate, and i need to remember to switch mindsets with processor speeds (forgot this isn't an rp2040)

main.c
trentgill: all the HW timers are assigned to the metro module (there's 8 available). i think it's best to leave them here, because the priority is allowed to slip if the CPU usage increases. As it is though, we're getting >4000 event loops per millisecond, so i think we're in good shape!

trentgill:
@tehn let me know once you can test on hardware (esp the i2c stuff) so i can merge this :)

tehn:
LED works now.

USB transactions aren't really visible on the discrete level, ie prints, probably because of multiple XORs being executed so the blip is super short. i didn't invest in tracking all the XOR usb transmission overlaps. setting input stream to 0.05 is very visual.

i reversed the times (long = connected, short = no usb) initially because i thought the discrete print/etc packets might be more visual, but that's not really helpful w/o fixing the XOR overlaps

tehn:
also--- the way things work right now--- if input stream is turned on and then USB is disconnected, it still keeps fast-usb-blinking.

trentgill:
@tehn thanks for this! i can make some quick changes this afternoon to fix the above issues. should be pretty quick, but understandably easier for me to change!

tehn:
honestly i don't think it's an issue-- could just be left as-is.

i don't have the right equipment to test i2c here, can do so later this evening if i can get out into town, or tomorrow.

PR #481, opened 2023-04-01
bytecode cross-compiled on dev machine

trentgill:
this PR is a pretty big refactor of how we run lua code on crow. previously we would save all the baked-in lua files in their regular string representation & load (compile) them at runtime on crow.

this PR includes an additional luacc (cross-compilation) stage which compiles the lua files into lua bytecode and saves that into the flash image. the benefit is much reduced runtime RAM when loading libraries as they can be loaded directly into the VM without needing to be parsed at runtime.

the RAM used by the lua machine is *not* decreased, nor is FLASH space decreased (bytecode is similar in size to the strings), but it's just the compilation process that doesn't need to allocate anything.

while this has improved the situation where crow would freeze when running scripts that utilize multiple different `ii` devices, it still hasn't solved the problem. previously we would fail on the 2nd `ii` device being loaded, and now we fail on the 3rd. it's also not a matter of the garbage collector needing to be tuned, as a full GC cycle is run after any file is loaded into the VM.

still, this feels like an improvement of the system, and i would like to push it into `main`. that said, there's no point rushing it out now as we still haven't solved the fundamental problem which led me to work on this.


tehn:
ok going in!!

tehn (approved):

PR #495, opened 2025-01-20
add missing wtape loop_active getter

midouest:
Reproduction:
```lua
> ii.wtape.event=function(e,v) print(e.name.."="..v) end

> ii.wtape[1].get("loop_active")

```

After the fix:
```lua
> ii.wtape.event=function(e,v) print(e.name.."="..v) end

> ii.wtape[1].get("loop_active")
loop_active=0.0

> ii.wtape[1].get("loop_active")
loop_active=1.0
```


tehn:
thank you!

PR #498, opened 2025-12-09
Clock features

evnoj:
Consider this as a draft.

I've added 3 features to the clock system. I'm happy to make changes, amend this PR to drop some of the features, or split this into multiple PRs. If you don't want these features, that's fine too.

For any features you do want to merge, I can also add the corresponding documentation to the [docs repo](https://github.com/monome/docs). I can also go through the steps to [make a release](https://github.com/monome/crow/blob/main/readme-development.md#making-a-release) if desired.

---
**`clock.sync` offset capability**

e2c23ff adds an optional `offset` parameter to `clock.sync()`, which is intended to work exactly as it does on norns. The implementation closely follows the [norns implementation](https://github.com/monome/norns/blob/main/matron/src/clocks/clock_scheduler.c#L43).

I figure you're familiar with how this works, but let me know if you'd like me to elaborate on this feature.

---
**`clock.handlers.tempo_change` user-defined handler**

ce0028b adds a handler that is called when the clock tempo changes, like `clock.tempo_change_handler` on norns. Since crow [modifies the `clock` metatable](https://github.com/monome/crow/blob/main/lua/clock.lua#L107) to prevent setting new indices, I took the approach of using the existing `clock.transport` table, but added a new key `clock.handlers` that points to this same table. We could simply have `clock.transport.tempo_change` to do away with the new `handlers` key. We could also set `clock.tempo_change_handler` to an empty function that would then always be called when the tempo changes, but this would allow the user to set `clock.tempo_change_handler = function(tempo) ... end` like on norns.

---
**`clock.time_since_last_input()` to check when an external clock trigger was last received**

b661e39 adds a function `clock.time_since_last_input()` which returns the time in seconds since a trigger was received at an input set to drive the clock by `input[n].mode('clock', div)`. If no input has been set to drive the clock, it returns -1.

I wanted to make a script that could be externally clocked, but only if an input was actually receiving a clock. Because there is no "jack detection", the only way I could think to do this is with a timeout that switches back to being unclocked if no external clock is received for some amount of time. I could not figure out a way to access the time that the last external trigger was received, although it is tracked internally in the C layer. I simply added a getter for this value. This enables a "timeout" like this:

```lua
function await_clock()
    input[2].mode( 'change', 3, 0.1, 'rising' )
    input[2].change = function()
        input[2].mode( 'clock', 1/4)
        print('clocked')
        clock_timeout:start()
    end
end

clock_timeout = metro.init{
    event = function()
        if clock.time_since_last_input() > 4 then -- 4 second timeout
            clock_timeout:stop()
            await_clock()
            unclocked()
        end
    end,
    time  = 1.0,
    count = -1
}

function unclocked()
    print('unclocked')
end

await_clock()
```
---

Let me know if you have questions, concerns, or want changes!

### DIFF
--- a/lib/clock.c
+++ b/lib/clock.c
@@ -99,13 +99,14 @@ bool clock_schedule_resume_sleep( int coro_id, float seconds )
     return ll_insert_event(&sleep_head, coro_id, wakeup);
 }
 
-bool clock_schedule_resume_sync( int coro_id, float beats ){
+bool clock_schedule_resume_sync( int coro_id, float beats, float offset ){
     double dbeats = beats;
 
     // modulo sync time against base beat
     double awaken = floor(reference.beat / dbeats);
     awaken *= dbeats;
     awaken += dbeats;
+    awaken += (double)offset;
 
     // check we haven't already passed it in the sub-beat & add another step if we have
     // we have to loop because fractional beats values may occur >2 times within a beat
--- a/lib/clock.c
+++ b/lib/clock.c
@@ -329,3 +329,11 @@ void clock_crow_in_div( float div )
 {
     crow_in_div = 1.0/div;
 }
+
+double clock_get_crow_last_time(void)
+{
+    if( clock_crow_last_time_set == false ){
+        return -1.0; // No clock received yet
+    }
+    return clock_get_time_seconds() - clock_crow_last_time;
+}
--- a/lib/clock.c
+++ b/lib/clock.c
@@ -8,6 +8,7 @@
 #include <stm32f7xx_hal.h> // HAL_GetTick
 #include "clock_ll.h" // linked list for clocks
 
+#include "caw.h" // Caw_printf
 
 ///////////////////////////////
 // private types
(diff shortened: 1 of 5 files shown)


tehn:
These all sound great, thank you! Will be able to check out the code soon.

b

-------- Original Message --------

evnoj:
I just realized I left an unneeded `#include "caw.h"` in lib/clock.c because I was using `Caw.printf` while testing. I'll remove it  in that commit with a force push during cleanup before merging, after we've decided on any other changes that need to be made.

tehn (approved):
this looks good to me. haven't tested on hardware.

maybe @trentgill will have a minute to glance at this, otherwise we can merge end of the week.

trentgill:
all looks good to me at a glance - i had no idea the `clock` lib had been extended on norns :/

won't have a chance to test this any time soon, but if @evnoj has done a good amount, then i'm happy to merge!

thanks for the work here!

PR #499, opened 2026-04-17
fix: clean up tempo change handler

evnoj:
In #498, I added `clock.handlers.tempo_change` as a user-defined callback when the tempo changes. I forgot to clean it up in `clock.cleanup`, so the callback was persisting through a `^^clear`. This fixes that bug.


tehn (approved):
thank you!

tehn:
@trentgill shall we call this v4.0.6?

